PR #13, opened 2021-05-14
pine64/pinebook-pro: Add cw2015 fuel gauge to device tree

zhaofengli:
This PR adds a device tree patch from mainline Linux for the cw2015 fuel gauge. With this added, the two device trees are mostly identical.


samueldr:
Ideally we wouldn't do that per-device. #14 is mostly about that.

But since we don't have an ideal solution yet, let's at least make the default DT more useful.

Thanks!

PR #22, opened 2021-05-23
Emit patches for a given Tow-Boot build

samueldr:
The outputs, which in turn are used for the release tarballs, now includes a `patches` directory with all the patches applied on top of U-Boot.

Combine this with the fact the `.config` file is available in the root of the tarball, it should be easier for individuals to re-combine a given Tow-Boot release on top of U-Boot without using the tooling here.

Note that this is not meant to replace the tooling. This is mainly to better handle GPL compliance by providing all patches side-by-side with the binaries.

### DIFF
--- a/support/builders/tow-boot/default.nix
+++ b/support/builders/tow-boot/default.nix
@@ -256,7 +256,7 @@ let
     } // meta;
 
     passthru = {
-      inherit patchset;
+      inherit mkOutput patchset;
     };
 
   } // removeAttrs args [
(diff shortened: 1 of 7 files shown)


support/builders/tow-boot/default.nix
```diff
@@ -64,201 +64,228 @@ let
...
     mkdir -p $out
-    cp .config $out
-    ${lib.optionalString (builtins.length filesToInstall > 0) ''
-    cp ${lib.concatStringsSep " " filesToInstall} $out
-    ''}
-    runHook postInstall
+    cd $out
+    ${lib.concatMapStringsSep "\n" (p: "cp ${p} ./${baseNameOf (toString p)}") tow-boot.patches}
+    cat <<EOF > series
+    ${lib.concatMapStringsSep "\n" (p: "${baseNameOf (toString p)}") tow-boot.patches}
```
samueldr, line 286: This will break if two patches end up named the same.

But it's not much of an issue as if two patches end up having the same exact name, it will break loudly since the `cp` will fail overwriting the file.

Issue #25, opened 2021-06-01
Document that setting the Legacy Bootable flag is needed for many distroboot boot options


samueldr:
E.g.

> If you use sgdisk, type
> ```
> sgdisk -A 3:set:2 /dev/sda
> ```
> 
> which means _set capability bit 2 (=the "legacy BIOS partition") of partition 3 of /dev/sda_.

— https://www.lightofdawn.org/wiki/wiki.cgi/BIOSBootGPT

b-:
Huh, I didn't know this was a thing. I've had trouble booting with CSM on an Intel Mac before, and I wonder if this had anything to do with it.

samueldr:
If it was, it was unrelated to this specific problem. (Being that CSM and UEFI and Intel Mac are not U-Boot platforms :))

Though yeah it's plausible that an "old-school" legacy boot bootloader (booted through CSM) looks at the "bootable" flag, which in GPT has to be set in a different manner.

b-:
Indeed, I just didn't even know there was a "legacy bootable" flag in GPT. I should read more about it

samueldr:
In theory if the software was unaware of GPT and the system was using an ["Hybrid MBR"](https://en.wikipedia.org/wiki/GUID_Partition_Table#Hybrid_MBR_(LBA_0_+_GPT)) it wouldn't use the legacy flag, it would read the bootable flag on the MBR.

Though that fact only serves to muddy the waters some more :).

PR #26, opened 2021-06-02
boards: Add olimex-teresI

samueldr:
*I do not own this board, it will need a tester before this is merged.*

### DIFF
--- a/boards/olimex/default.nix
+++ b/boards/olimex/default.nix
@@ -1,0 +1,5 @@
+{ Tow-Boot, allwinnerA64 }:
+
+{
+  olimex-teresI = allwinnerA64 { defconfig = "teres_i_defconfig"; };
+}


Thra11:
I created an SD card with the Tow-Boot disk-image, followed by two more partitions, a 20M firmware partition (empty except for a single file, `config.txt`), and a bootable system partition containing a nixos system (These two partitions are copied from another SD card with a working system using u-boot). The internal eMMC does not currently have any bootloaders or bootable partitions on it.

If I fit this SD card, power on the system, and leave it alone, it boots just fine.

If I interrupt it by pressing Ctrl+C when prompted, it shows the menu, which is navigable using the keyboard. However, all of the "Boot from *" options fail, even "Boot from SD", despite there being a (clearly) bootable SD card fitted. Will continue to investigate...

samueldr:
~~A fresh build from the current tip of `development` or using that first release? It's possible [cancelling boot with ^C is cancelling the commands list](https://github.com/Tow-Boot/Tow-Boot/commit/5ec6abbe8c85758084d27d0a5c3179e08d52a978) if you used the first release.~~ DUH... It's a build from this PR... disregard this.

Note that the "firmware partition" you're adding is useless; in the NixOS generic sd image, that "FIRMWARE" partition is the IBF (Initial Boot Firmware, read: U-Boot) exclusively used for the Raspberry Pi family of hardware.

Is that "bootable system partition" partition marked as *bootable*? Starting from the `disk-image.img` file it would be GPT. See #25 for some notes about setting a GPT partition as "bootable".

I guess I could test on any of my A64 targets for this. It wouldn't surprise me if the *Boot from __* options are failing due to some weirdness, impurities and environmental pollution. I might have seen a similar issue previously while trying to boot some targets.

Thra11 (approved):
Have logged the problems with the menu as #29, which seems unlikely to be specific to this board, so looks good to me.

Thra11:
> Is that "bootable system partition" partition marked as _bootable_?

Yes. I added a legacy bootable flag in fdisk.

> Note that the "firmware partition" you're adding is useless.

I believe you're probably right. In theory the `config.txt` is supposed to be read to set `debug=on`, in order to enable/disable the headphone jack serial output (It doesn't have a physical switch like the pinebook-pro and pinephone). It's unlikely to be doing anything at the moment, just lying around on a random partition.

PR #30, opened 2021-06-11
Add support for the ODROID N2

L-as:
I've tested this with [nixos.iso_minimal_new_kernel.aarch64-linux](https://hydra.nixos.org/job/nixos/trunk-combined/nixos.iso_minimal_new_kernel.aarch64-linux) and it works fine, except for the fact that u-boot doesn't support writing to the SPI currently.

Like the C2 it needs some special considerations and therefore doesn't have a separate builder.

I haven't removed building the spi-installer.img, since it will likely work in the future, but should I remove it?

### DIFF
--- a/boards/odroid/default.nix
+++ b/boards/odroid/default.nix
@@ -11,6 +11,11 @@
       # ODROID N2 SPI support
       ./0001-Enable-the-SPI-on-the-ODROID-N2-by-default.patch
     ];
+    extraConfig = ''
+      CONFIG_USE_PREBOOT=y
+      CONFIG_PREBOOT="usb start ; usb info"
+    '';
+
   };
   odroid-C2 = Tow-Boot.systems.aarch64.callPackage ./odroid-c2.nix { };
 }
--- a/boards/odroid/default.nix
+++ b/boards/odroid/default.nix
@@ -1,5 +1,18 @@
-{ Tow-Boot }:
+{ Tow-Boot, amlogicG12, amlogicFirmware }:
 
 {
+  odroid-N2 = amlogicG12 {
+    boardIdentifier = "odroid-N2";
+    defconfig = "odroid-n2_defconfig";
+    FIPDIR = "${amlogicFirmware}/odroid-n2";
+    # Access to the SPI flash itself works fine, but the SPI installer menu
+    # doesn't work, shows visual artifacts, and somehow gets stuck in that state.
+    withSPI = false;
+    SPISize = 8 * 1024 * 1024;
+    patches = [
+      # ODROID N2 SPI support
+      ./0001-Enable-the-SPI-on-the-ODROID-N2-by-default.patch
+    ];
+  };
   odroid-C2 = Tow-Boot.systems.aarch64.callPackage ./odroid-c2.nix { };
 }
--- a/boards/odroid/default.nix
+++ b/boards/odroid/default.nix
@@ -10,6 +10,8 @@
     patches = [
       # ODROID N2 SPI support
       ./0001-Enable-the-SPI-on-the-ODROID-N2-by-default.patch
+      # fix DTB file location on N2+. to be removed once we upgrade to U-Boot 2021.10
+      ./0001-Fix-n2-plus-fdtfile.patch
     ];
     extraConfig = ''
       CONFIG_USE_PREBOOT=y
--- a/boards/odroid/default.nix
+++ b/boards/odroid/default.nix
@@ -5,9 +5,7 @@
     boardIdentifier = "odroid-N2";
     defconfig = "odroid-n2_defconfig";
     FIPDIR = "${amlogicFirmware}/odroid-n2";
-    # Access to the SPI flash itself works fine, but the SPI installer menu
-    # doesn't work, shows visual artifacts, and somehow gets stuck in that state.
-    withSPI = false;
+    withSPI = true;
     SPISize = 8 * 1024 * 1024;
     patches = [
       # ODROID N2 SPI support
--- a/support/builders/amlogic-g12/default.nix
+++ b/support/builders/amlogic-g12/default.nix
@@ -45,12 +45,14 @@ let
     ];
 
     postBuild = ''
+      echo " :: Merging with proprietary components..."
+      (PS4=" $ "; set -x
       meson64-pkg --type bl30 --output bl30.pkg $FIPDIR/bl30.bin $FIPDIR/bl301.bin
       meson64-pkg --type bl2 --output bl2.pkg $FIPDIR/bl2.bin $FIPDIR/acs.bin
       meson64-bl30sig --input bl30.pkg --output bl30.30sig
       meson64-bl3sig  --input bl30.30sig --output bl30.3sig
       meson64-bl3sig  --input $FIPDIR/bl31.img --output bl31.3sig
-      meson64-bl3sig  --input Tow-Boot.bin --output bl33.3sig
+      meson64-bl3sig  --input u-boot.bin --output bl33.3sig
       meson64-bl2sig  --input bl2.pkg --output bl2.2sig
       args=(
         --bl2 bl2.2sig
@@ -69,8 +71,8 @@ let
       test -e $FIPDIR/aml_ddr.fw && args+=(--ddrfw8 $FIPDIR/aml_ddr.fw)
       test -e $FIPDIR/lpddr3_1d.fw && args+=(--ddrfw9 $FIPDIR/lpddr3_1d.fw)
 
-      meson64-bootmk --output Tow-Boot2.bin "''${args[@]}"
-      mv Tow-Boot2.bin Tow-Boot.bin
+      meson64-bootmk --output Tow-Boot.bin "''${args[@]}"
+      )
     '';
 
     installPhase = ''
--- a/support/builders/amlogic-g12/default.nix
+++ b/support/builders/amlogic-g12/default.nix
@@ -1,0 +1,94 @@
+{ lib, buildTowBoot, meson64-tools, imageBuilder, runCommandNoCC, spiInstallerPartitionBuilder }:
+
+let self =
+{ FIPDIR, withSPI ? false,  ...} @ args':
+
+let
+  args = removeAttrs args' (builtins.attrNames (builtins.functionArgs self));
+
+  partitionOffset = 1;
+  sectorSize = 512;
+  firmwareMaxSize = imageBuilder.size.MiB 4;
+
+  firmwarePartition = imageBuilder.firmwarePartition {
+    inherit sectorSize;
+    partitionOffset = partitionOffset; # in sectors
+    partitionSize = firmwareMaxSize; # in bytes
+    firmwareFile = "${firmware}/binaries/Tow-Boot.noenv.bin";
+  };
+
+  baseImage' = extraPartitions: imageBuilder.diskImage.makeMBR {
+    name = "disk-image";
+    diskID = "01234567";
+    # Align on 512 sector sizes.
+    # This is not for the actual partitions for the end-user, only so we
+    # are able to make the first partition start at sector 0x01
+    alignment = 512;
+
+    partitions = [
+      firmwarePartition
+    ] ++ extraPartitions;
+  };
+  baseImage = baseImage' [];
+  spiInstallerImage = baseImage' [
+    (spiInstallerPartitionBuilder {
+      firmware = firmwareSPI;
+    })
+  ];
+
+  firmware' = variant: buildTowBoot ({
+    meta.platforms = [ "aarch64-linux" ];
+    inherit variant FIPDIR;
+
+    nativeBuildInputs = [
+      meson64-tools
+    ];
+
+    postBuild = ''
+      meson64-pkg --type bl30 --output bl30.pkg $FIPDIR/bl30.bin $FIPDIR/bl301.bin
+      meson64-pkg --type bl2 --output bl2.pkg $FIPDIR/bl2.bin $FIPDIR/acs.bin
+      meson64-bl30sig --input bl30.pkg --output bl30.30sig
+      meson64-bl3sig  --input bl30.30sig --output bl30.3sig
+      meson64-bl3sig  --input $FIPDIR/bl31.img --output bl31.3sig
+      meson64-bl3sig  --input Tow-Boot.bin --output bl33.3sig
+      meson64-bl2sig  --input bl2.pkg --output bl2.2sig
+      args=(
+        --bl2 bl2.2sig
+        --bl30 bl30.3sig
+        --bl31 bl31.3sig
+        --bl33 bl33.3sig
+
+        --ddrfw1 $FIPDIR/ddr4_1d.fw
+        --ddrfw2 $FIPDIR/ddr4_2d.fw
+        --ddrfw3 $FIPDIR/ddr3_1d.fw
+        --ddrfw4 $FIPDIR/piei.fw
+        --ddrfw5 $FIPDIR/lpddr4_1d.fw
+        --ddrfw6 $FIPDIR/lpddr4_2d.fw
+        --ddrfw7 $FIPDIR/diag_lpddr4.fw
+      )
+      test -e $FIPDIR/aml_ddr.fw && args+=(--ddrfw8 $FIPDIR/aml_ddr.fw)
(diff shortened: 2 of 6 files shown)


support/overlay/overlay.nix
```diff
@@ -63,6 +63,15 @@ in
...
 
+    meson64-tools = let
+      src = final.fetchFromGitHub {
+        owner = "angerman";
+        repo = "meson64-tools";
+        rev = "a2d57d11fd8b4242b903c10dca9d25f7f99d8ff0";
+        sha256 = "Wyx4ngfDN6jSEBwrgH8z44wntJEKMuCQz56MrU9mB5E=";
+      };
+    in (import src { nixpkgs = final; }).meson64-tools;
+
```
samueldr: We shouldn't do IFDs; they'll cause issues when we get some CD done through Hydra.

samueldr:
> Like the C2 it needs some special considerations and therefore doesn't have a separate builder.

Are you sure it needs special consideration?

The S922X is G12B; and as I shared on the matrix channel, we have a path forward for a generic G12A/G12B packaging.

 - https://github.com/NixOS/nixpkgs/blob/f02c05e056b9bef1b8ebc445a3046432a68d4314/pkgs/misc/uboot/amlogic.nix#L28-L29 (From https://github.com/NixOS/nixpkgs/pull/119371)

AFAICT the build is identical, save that the `meson64-tool` are prefixed in the Nixpkgs build (since the names are a bit too generic).

It should be okay to add an amlogicG12 builder, which applies to G12A and G12B.

L-as:
Thanks, I didn't know that, I'll try fixing both of these issues.

L-as:
I fixed the issues, and access to the SPI flash works fine, but the installer image somehow gets stuck in the menu. Because of that I disabled building the installer image for it.

support/overlay/overlay.nix
```diff
@@ -63,6 +63,13 @@ in
 
     gxlimg = callPackage ./gxlimg { };
 
+    meson64-tools = let
+      src = builtins.fetchTarball {
+        url = "https://github.com/angerman/meson64-tools/archive/a2d57d11fd8b4242b903c10dca9d25f7f99d8ff0.tar.gz";
+        sha256 = "1487cr7sv34yry8f0chaj6s2g3736dzq0aqw239ahdy30yg7hb2v";
+      };
+    in (import src { nixpkgs = final; }).meson64-tools;
+
```
samueldr: This is still an IFD; we'll need to package it locally with a proper `callPackage` use.

support/builders/tow-boot/default.nix
```diff
@@ -99,6 +99,9 @@ let
 
       # Intrusive opinionated patches
       ./patches/0001-Tow-Boot-sunxi-ignore-mmc_auto-force-SD-then-eMMC.patch
+
+      # ODROID N2 SPI support
+      ./patches/0001-Enable-the-SPI-on-the-ODROID-N2-by-default.patch
```
samueldr: Board-specific patches live in the board vendor's directory, and are included only for the given board.

Additionally we might want to include this patch only when relevant, as discussed.

support/builders/amlogic-g12/default.nix
```diff
@@ -0,0 +1,74 @@
...
+  };
+  baseImage = baseImage' [];
+  spiInstallerImage = baseImage' [
+    (spiInstallerPartitionBuilder {
+      inherit defconfig;
+      firmware = "${firmware}/u-boot.bin";
+    })
+  ];
+
+  # Perhaps this could be shared with other boards and be refactored into a shared builder.
```
samueldr:
```suggestion
```

:)

support/overlay/overlay.nix
```diff
@@ -83,7 +90,11 @@ in
     };
 
     amlogicGXL = aarch64.callPackage ../builders/amlogic-gxl {
-      gxlimg = final.Tow-Boot.gxlimg;
+      inherit (final.Tow-Boot) gxlimg;
```
samueldr, line 88: Please limit unrelated fixes and changes to specific discrete commits. The fixes like this one *are welcome*.

Otherwise it gets harder to `git blame` and sometimes even `git bisect` through unrelated changes.

samueldr (changes requested):
Some changes needed.

In addition I'd like to see here discrete commits for the logical changes.

 - Minor fix to overlay.nix (the inherit thing)
 - Introduction of meson64-tools
 - Introduction of the `amlogicG12` builder
 - Introduction of the odroid-N2 board

support/overlay/overlay.nix
L-as: I don't think it actually is. For one, it builds with --no-allow-import-from-derivation, and in addition, fetchTarball doesn't produce a derivation, it produces the result directly (even though it might possibly create a derivation for it internally).

support/builders/tow-boot/default.nix
L-as: Isn't it always relevant in the u-boot stage though? Wouldn't it be convenient always to have access to the SPI in the boot stage? Not like you do any intensive computations in Tow-Boot, so I don't think the half eMMC bandwidth is that relevant.

support/overlay/overlay.nix
samueldr: Oh... right... It's technically not an IFD, but we still want to package it ourselves.

Please use this, as it's intended to go in Nixpkgs upstream at some point, and eventually we'll drop our packaging of it.

 - https://github.com/NixOS/nixpkgs/blob/f02c05e056b9bef1b8ebc445a3046432a68d4314/pkgs/misc/meson64-tools/default.nix

The tool names in the builder should also use the `meson64-*` qualified aliases this adds. To ensure the generic names are not trampled by other tools. This is because *if* a tool were to trample it in the future, that would be silent.

support/builders/tow-boot/default.nix
samueldr: Yeah, "might" want. Not sure. You're probably right.

support/builders/amlogic-g12/default.nix
```diff
@@ -0,0 +1,74 @@
...
+      bl3sig  --input u-boot.bin --output bl33.3sig
+      bl2sig  --input bl2.pkg --output bl2.2sig
+      args=()
+      fws=(ddr4_1d.fw ddr4_2d.fw ddr3_1d.fw piei.fw lpddr4_1d.fw lpddr4_2d.fw diag_lpddr4.fw aml_ddr.fw lpddr3_1d.fw)
+      for i in "''${!fws[@]}"; do
+        fw="$FIPDIR/''${fws[i]}"
+        test -e $fw && args+=("--ddrfw$((i+1))" $fw)
+      done
+      bootmk --output u-boot2.bin --bl2 bl2.2sig --bl30 bl30.3sig --bl31 bl31.3sig --bl33 bl33.3sig "''${args[@]}"
+      mv u-boot2.bin u-boot.bin
```
samueldr: I'm pretty sure I'm being rude right here, but please bear with me :)

I would prefer if this segment from the Nixpkgs amlogic PR was used outright:

 - https://github.com/NixOS/nixpkgs/blob/f02c05e056b9bef1b8ebc445a3046432a68d4314/pkgs/misc/uboot/amlogic.nix#L36-L88

After all, I wrote it entirely considering it would be used in Tow-Boot once I had someone handy to test and validate it works.

While it's mostly stylistic differences, there is one huge difference: `ddrfwX` files are used specifically by name and index, rather than *tried*. like you do here. While your approach is probably better in many ways, I'm not able to confirm that it is good. My main concern is that if for a board one of the firmware from 1 through 7 is missing, we want to break. We want to break as we are not expecting that any of them are missing *going on from all the current knowledge*.

samueldr:
Sorry I'm quite nitpicky here :). Especially given it's something I had partially done elsewhere, so I have a lot of preconceived notions.

Otherwise, it looks alright mainly.

support/builders/amlogic-g12/default.nix
L-as: What board doesn't need ddrfw8?

samueldr: Compare those listings:

 - https://github.com/LibreELEC/amlogic-boot-fip/tree/master/odroid-n2
 - https://github.com/LibreELEC/amlogic-boot-fip/tree/master/odroid-hc4

This is for one vendor.


Then another sample:

 - https://github.com/LibreELEC/amlogic-boot-fip/tree/master/sei510
 - https://github.com/LibreELEC/amlogic-boot-fip/tree/master/sei610

So it's definitely not unexpected to have only those two missing. Furthermore IIRC LibreELEC also has a similar (but differently implemented) logic, where ddrfw8/9 end up being conditional on the files existing. And all fw files are at *specific numbered offsets*

L-as:
It doesn't build now for some reason after updating the branch. I get this when building u-boot:
```
  SHIPPED dts/dt.dtb
  CAT     u-boot-dtb.bin
  COPY    u-boot.dtb
  COPY    u-boot.bin
  LD      u-boot.elf
```

samueldr:
I always have trouble remembering how to push to a PR directly, so here's the patch to fix this:

 - https://gist.github.com/514fb6cdb922918a8c9eabbaaad3cbf3

This also adds some `set -x` magic to better help point out which command fails. This is how I found out what failed.

tpwrules:
I found that if I add `CONFIG_USE_PREBOOT=y` to `extraConfig`, then the USB keyboard works in the SPI flasher and it runs.

However, there is a bizarre 5 second delay on some screen updates, unless a key is pressed, when a keyboard is attached. I haven't been able to nail down what the problem is yet, or if it's Tow-Boot's fault.

samueldr:
> I found that if I add `CONFIG_USE_PREBOOT=y` to `extraConfig`, then the USB keyboard works in the SPI flasher and it runs.

I suggested it to you yesterday, though I guess we did talk about lots of things.

> However, there is a bizarre 5 second delay on some screen updates, unless a key is pressed, when a keyboard is attached. I haven't been able to nail down what the problem is yet, or if it's Tow-Boot's fault.

A not-so-bad way to test for this would be to get a distro that uses GRUB with the **console** driver (NixOS uses the graphical driver). When GRUB is using the console driver, it uses the same codepaths U-Boot itself uses to render the console. So I assume it's likely it would reproduce the issue there. Do reproduce on Tow-Boot+GRUB first before trying stock U-Boot+GRUB.

L-as:
I fixed the issues mentioned, but haven't tested it yet.

Edit: It *does* build.

samueldr:
Oh, we discussed a bit on the matrix room for NixOS on ARM, and we figured out something to investigate about the broken graphics,

IIRC, it looked like @tpwrules was able to remove all corruption, which really is the interface hanging by using `usb info` after `usb start`. Though at the very least, when the interface hangs, typing more on the keyboard "unclogs" something (if you allow the figure of speech), and it ends up being able to finish redrawing.

As I don't have the hardware, I can't test more, or investigate, but this points to device-specific USB issues. It is unclear whether the specific USB keyboards may be at fault or not.

L-as:
It's probably a hardware bug in the ODROID N2. It has a bunch of other USB issues.

boards/odroid/default.nix
```diff
@@ -1,5 +1,18 @@
...
+{ Tow-Boot, amlogicG12, amlogicFirmware }:
 
 {
+  odroid-N2 = amlogicG12 {
+    boardIdentifier = "odroid-N2";
+    defconfig = "odroid-n2_defconfig";
+    FIPDIR = "${amlogicFirmware}/odroid-n2";
+    # Access to the SPI flash itself works fine, but the SPI installer menu
+    # doesn't work, shows visual artifacts, and somehow gets stuck in that state.
+    withSPI = false;
```
tpwrules: You should be able to turn SPI back on and remove the comment.

L-as: Done.

tpwrules:
I found that if I add `CONFIG_PREBOOT="usb start ; usb info"` for the N2 build then the USB keyboard works fine. Alternately, if you can manage to get into the firmware console, simply running the `usb info` command fixes the problem, at least until `usb reset`. We may want to add `usb info` to the "Rescan USB" menu command because otherwise running it will break the keyboard and confuse users.

I've tested the SPI flasher, and the firmware once flashed, on my ODROID-N2 Plus revision 0.6 4GB. It seems to work great. I can flash the firmware, then run the flashed firmware from SPI and boot from a USB drive into UEFI GRUB.

As for the keyboard issue, affected keyboards seem to have a 5 second time delay to recognize input, or another key must be pressed to dislodge the first key. If you can get into the firmware console with one, which requires some patience, running the `help` command will produce one line of output every 5 seconds, or one line each time a key is pressed or released. It's present in upstream U-Boot 21.04 and 21.10, as well as the latest Amlogic-specific fork.

All the problematic keyboards I've tested seem to have internal USB 1.1 hubs which the keyboard is connected through. Keyboards with USB 2.0 hubs, or external USB 2.0 hubs don't have problems. The speed of the keyboard itself doesn't seem to matter. I may see if I can dig up a USB 1.1 hub to test a working keyboard through, but those may be pretty difficult to find outside of 1998.

L-as:
@tpwrules Do you suggest that I do this for just the N2 or all G12s?

tpwrules:
I would imagine just to the N2 for now.

Then it should be ready for merging!

L-as:
I don't think there are any remaining problems, but it hasn't been tested on my personal hardware. AFAICT @tpwrules has tested it, so it ought to be find?

samueldr:
I would like a confirmation that at the current state of this branch, without any additional changes, it does work correctly. Just in case. I don't have the hardware to test.

But yes, it seems like it will be fine.

L-as:
Unfortunately, I still get nothing responsive in the SPI installer. No USB device turns on, and I'm not sure what's happening. This is exactly what happened for me before too. Does it work for you? @tpwrules

Maybe my ODROID is just faulty.

samueldr:
@L-as just checking, you know that the USB support in U-Boot (and thus Tow-Boot) is not hot-plug friendly? Here the keyboard would need to be already plugged-in when `CONFIG_PREBOOT` is running. Which is early enough during boot that we might as well say that it needs to be plugged-in before powering on.

L-as:
It was plugged in from the start @samueldr

tpwrules:
I just re-tested everything from scratch. Here is what I did to build it:
```
git clone https://github.com/Tow-Boot/Tow-Boot
cd Tow-Boot
curl 'https://patch-diff.githubusercontent.com/raw/Tow-Boot/Tow-Boot/pull/30.patch' | git apply -
NIXPKGS_ALLOW_UNFREE=1 nix-build -A odroid-N2
sudo dd if=result/spi.installer.img of=/dev/<an SD card> conv=notrunc,sync
```

Here is what I did to test it:
1. Insert SD card into ODROID, remove eMMC module, and set switch to "MMC"
2. Unplug all USB devices, then plug in keyboard and HDMI
3. Plug power into ODROID and wait for boot menu to show up on monitor
4. Verify that keyboard does not behave weirdly, select "Flash firmware to SPI", wait for process to complete
5. When flash is finished and it prompts to reboot, unplug power, remove SD card, and set switch to "SPI"
6. Plug power back in, wait for boot menu to show up, verify that keyboard does not behave weirdly

Everything worked as expected in the bootloader, although it couldn't boot my NixOS install off another SD card. To fix that problem, I had to add one last patch: https://gist.github.com/tpwrules/61c0b2c41d3edd5ef4b0a7e15f1a462b

It's based off this upstream commit, although not identical: https://source.denx.de/u-boot/u-boot/-/commit/480aeb0aa7b1fcb78e15d1678610f61ff421b883

With that patch, I re-did all the above steps (now including applying that patch in the gist), and everything still worked as described and NixOS booted off my SD card successfully.

L-as:
Oh, you're using an N2+ then?

tpwrules:
Yes, it's an ODROID-N2+. It says "Rev 0.6 20210121". Are the CPUs different in any aspect except clock speed? Did they fix any errata or whatever?

L-as:
@tpwrules Pretty sure it has a bunch of fixes. I added your patch. I think this is ready for merging?

tpwrules:
Yup, I double checked and the patches applied correctly. I'd say it's ready, modulo whatever the thoughts are on it potentially being broken for regular N2.

Did you re-test on your board according to my instructions? Can you try different keyboards?

L-as:
I already did it the way you described, so there's nothing to retest. I think my ODROID is just broken, it has some other problems too.

samueldr (approved):

PR #34, opened 2021-06-23
Take control over the environment

### LINKED ISSUE #19: Performing `saveenv` when booting from SPI corrupts SPI
This appears to be somewhat of [a u-boot problem](https://forums.xilinx.com/t5/Embedded-Linux/Where-does-U-Boot-saveenv-to/td-p/532915), and not specifically tow-boot, but I figure an opinionated configuration might provision for this sort of thing:

I had Tow-Boot installed to SPI, and I was poking around the console, and I tried running `saveenv`, which proceeded to save environment settings to SPI, overwriting part of the loader. Which part? I don't know. It didn't say _where_ in SPI it was saving, but I presume it was an important region because now the power LED is red upon power on and nothing else happens.

Maybe `saveenv` can write to a specific offset past the bootloader? I'm not sure what the best course of action is here. All I know is that I rendered my Pinebook Pro unbootable through a command with a relatively innocuous-sounding name. (Yes, I should have looked up what this did first…)

samueldr:
This is a weird name. But saying it "fixes `saveenv` issues" is too diminutive. But hey, closes #19!

* * *

### Environment

Two broad strokes. For release `002` of Tow-Boot, we will **only** support environment in the SPI Flash for builds installed to the SPI Flash.

Firmware installed to a shared storage media, starting with `002` up to later will not allow saving the environment.

To reduce confusion, a Tow-Boot install will **strictly** only use the environment in the same storage it is installed into. This means that SPI-installed firmware will read/save to the SPI Flash. Not implemented yet, but using the shared storage strategy, **only when installed to a protective partition** will Tow-Boot read/save the environment.

### Variants

Since we need to change how the firmware acts depending on where it is installed, it was deemed the easiest method to *just* build different binaries. So we're now producing "variants", which in turn affect the configuration slightly. These are somewhat different from "simple" feature flags, as we are changing a fundamental property of the firmware.

It's not expected we'll gain many more variants. In fact the only one I expect to add is `mmc` which (using the U-Boot nomenclature) will use the partition from the shared storage media to manage the environment.

In other words, variants right now are strictly dealing with where the environment is read and saved.

### File names and structure

It was coming, but since we needed to change file names for variants, I'm implementing the changes here. All files now report as "Tow-Boot". It apparently caused mild confusion in some users that the files were named "u-boot". Additionally, files are now stored in directories, `config` for the variant configuraiotns, and `binaries` for the "irrlevant to most users" files.

The main stars, the disk images, are at the root of the result.

* * *

### Environment with shared storage

As stated previously "**only when installed to a protective partition** will Tow-Boot read/save the environment."

This requires some development. There are no environment saving strategies that allows this to work in a safe manner.

Since we want to save to the same partition Tow-Boot is installed into, we need the environment "in MMC". With that strategy, *only* an offset to a *hardcoded* mmc device is supported. This means that using Tow-Boot on an SD card would save the environment to an arbitrary offset on the internal MMC storage, which **is likely to break things**! So we'll need to track the mmc device the system booted from. Additionally, it would be better if we'd instead be able to work with partitions, and offsets internal to the partition, instead of raw device offsets. This way it is more likely we don't need to implement platform-specific differences.

Using the "in FAT" strategy for the Raspberry Pi family of hardware might work... Except that it is unclear what happens when booting from USB. It will require extensive testing.

### DIFF
--- a/boards/libreComputer/default.nix
+++ b/boards/libreComputer/default.nix
@@ -24,6 +24,7 @@ in
     defconfig = "libretech-ac_defconfig";
     FIPDIR = "${amlogicFirmware}/lafrite";
     withSPI = true;
+    SPISize = 128 /* Mbits */ * 1024 * 1024 / 8; # equiv to 16 MiB
   };
   libreComputer-rocRk3399Pc = rocRk3399Pc {
     defconfig = "roc-pc-rk3399_defconfig";
--- a/boards/libreComputer/default.nix
+++ b/boards/libreComputer/default.nix
@@ -16,6 +16,7 @@ let
       substituteInPlace include/tow-boot_env.h \
         --replace 'setup_leds=echo\0' 'setup_leds=${setup_leds}\0'
     '';
+    SPISize = 128 /* Mbits */ * 1024 * 1024 / 8; # equiv to 16 MiB
   };
 in
 {
--- a/boards/odroid/odroid-c2.nix
+++ b/boards/odroid/odroid-c2.nix
@@ -26,7 +26,7 @@ let
     inherit sectorSize;
     partitionOffset = partitionOffset; # in sectors
     partitionSize = firmwareMaxSize; # in bytes
-    firmwareFile = "${firmware}/u-boot-combined.bin";
+    firmwareFile = "${firmware}/binaries/Tow-Boot.noenv.bin";
   };
 
   baseImage = baseImage' [];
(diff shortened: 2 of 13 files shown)


samueldr:
I think this breaks `uBoot-qemuX86_64`... need to check.

Looks like it needs environment

# TODO

 - [x] Look into `uBoot-qemuX86_64`

PR #35, opened 2021-06-23
raspberryPi-*: Increase malloc pool to 64 MiB (+env)

samueldr:
This fixes the issue where the logo wouldn't show on the Raspberry Pi family of hardware.

* * *

This is not really an upstreamable change. Really this probably shouldn't be hardcoded per platform in header files, but rather some `Kconfig` option?


samueldr:
Found the reason.

 - https://source.denx.de/u-boot/u-boot/-/blob/v2021.04/Kconfig#L211-217
 - https://source.denx.de/u-boot/u-boot/-/blob/v2021.04/include/configs/rpi.h#L57
 - https://source.denx.de/u-boot/u-boot/-/blob/v2021.04/include/configs/sunxi-common.h#L141-142
 - https://source.denx.de/u-boot/u-boot/-/blob/v2021.04/include/configs/meson64.h#L35

Basically the Raspberry Pi family of hardware uses a tiny amount of space for malloc (4MiB). This probably is *fine* and by design due to the considerations listed beforehand. The early models with 256MiB of RAM, in which a chunk is reserved for the videocore chip.

We probably can and should look into fulfilling the TODO, and conditionally apply different sizes depending on the target Raspberry Pi. Though we might also just want to increase it outright?

## TODO

 - [x] Re-do by increasing `CONFIG_SYS_MALLOC_LEN`

PR #44, opened 2021-06-29
doc: Add Arch Linux ARM startup guide

samueldr:
Ideally this would be managed by the upstream project, and we'd link to it.

Note that I only followed these instructions **once** while I was writing them. It's possible there's a silly mistake or more.

And it is **by design** that we're not installing a package to manage Tow-Boot or U-Boot. The extlinux.conf file is enough.

Oh, and the partition scheme? GPT is strongly suggested. There's no reasons, in 2021, to go with MBR, *except* Amlogic boards without SPI Flash. Or misplaced nostalgia? And only one partition? Makes things simpler. But advanced users will know what to do, hopefully, to make it work with more partitions.

Note that I am *unsure* about how the `parted` command will react with an SD card started from the shared disk image. I'll want to test:

 - [ ] Test with a shared disk image

It's not obvious how to tell `parted`, in the imperative command-based interface, "make one partition after the current one, fill the space".

**Verified to boot on**:

 -  Pine A64-LTS
 -  (partially) on Libre Computer ROC-RK3399-PC (but fails due to fusb302, built-in module, kernel bug where it kills the type-c power during boot)
 - *something* happens on the Pinebook Pro, but I was unable to determine the actual boot process as the display turned off. And I wasn't setup to debug it at the time.

So I guess it's YMMV for Pinebook Pro users.

### DIFF
--- a/doc/distributions/archlinuxarm.md
+++ b/doc/distributions/archlinuxarm.md
@@ -1,0 +1,157 @@
+Arch Linux ARM
+==============
+
+This guide uses the [*generic AArch64 installation*](https://archlinuxarm.org/platforms/armv8/generic) as a base.
+
+It is assumed that Tow-Boot has already been installed (e.g. to SPI), or that
+the storage media was initialized with the *shared storage* disk image. Please
+refer to [Getting Started](../getting-started.md).
+
+This installation guide assumes the use of the tarball extracted to a single
+non-encrypted root filesystem.
+
+This installation boots using the "extlinux.conf-compatible" boot scheme.
+
+Deviating from the usual setup, this uses only one partition for the system.
+The firmware is able to look in an EXT4 partition. Users are free to customize
+the setup as they see fit.
+
+
+Installation steps
+------------------
+
+> **NOTE**: This guide uses `/dev/sdX` as a generic placeholder for the target
+> storage device, and `/mnt` for the temporary mount point.
+>
+> Usage of `sudo` commands is used to denote commands requiring root
+> privileges.
+
+### 0. Preparations
+
+Download the latest generic tarball.
+
+```
+ ~ $ wget http://os.archlinuxarm.org/os/ArchLinuxARM-aarch64-latest.tar.gz
+```
+
+### 1. Partitions and filesystems
+
+#### 1.1. Initializing the disk
+
+> **WARNING**: Skip this step and go to (1.2)  if you are using the *shared
+> storage strategy*. The disk image will provide you with a usable and
+> pre-configured partition table.
+
+```
+ ~ $ sudo parted /dev/sdX -- mklabel gpt
+```
+
+#### 1.2. Adding partitions
+
+This adds the root partition, filling the drive.
+
+```
+ ~ $ sudo parted /dev/sdX -- mkpart primary archlinuxarm-rootfs 0 100%
+```
+
+Verify the number of the added partition. This guide assumes this is partition
+1, `/dev/sdX1` will be used to refer to this partition.
+
+```
+ ~ $ sudo parted /dev/sdX -- print
+Model: SD SD16G (sd/mmc)
+Disk /dev/sdX: 15.5GB
+Sector size (logical/physical): 512B/512B
+Partition Table: gpt
+Disk Flags: 
+
+Number  Start   End     Size    File system  Name                 Flags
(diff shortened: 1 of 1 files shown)


n2qz:
Partition step should be:
```
parted /dev/sdb -- mkpart archlinuxarm-rootfs 1 100%
```

qiu-x:
@n2qz Why can't the partition start at 0% when using the dedicated storage method?

n2qz:
> @n2qz Why can't the partition start at 0% when using the dedicated storage method?

Partition will end up misaligned if you try that (and parted will provide a warning to that effect).

benthetechguy:
Nice, I unknowingly wrote very similar instructions on the [Pine64 Wiki](https://wiki.pine64.org/wiki/Installing_Arch_Linux_ARM_On_The_Pinebook_Pro) for using Tow-Boot to get Arch Linux ARM running on the Pinebook Pro.

benthetechguy:
As for the last part, normal UEFI installation via pacstrap should be relatively simple except for the fact that you need an existing Arch Linux ARM install to run it from since pacstrap doesn't support foreign architectures like debootstrap.
Speaking of Debian, I have [a package](https://mentors.debian.net/package/pacman-package-manager) waiting for a Debian member to sponsor the upload of that would allow someone to package `pacstrap` for Debian, which would greatly simplify things, as many people can run some form of Debian easily on their boards.

EDIT: I just tested installing Arch Linux ARM via the official pacstrap installation method on the Arch Wiki and it worked perfectly without my needing to change anything except for the name of the kernel from `linux` to `linux-aarch64`. Even down to the partitioning and bootloader, everything worked exactly the same as it would on an x86 UEFI system. The only issue remaining is getting to an environment where pacstrap can be run.

benthetechguy:
> you need an existing Arch Linux ARM install to run it from since pacstrap doesn't support foreign architectures like debootstrap

> The only issue remaining is getting to an environment where pacstrap can be run.

OK, I have now figured out how to pacstrap Arch Linux ARM from x86_64 Arch Linux.

1. Install `qemu-user-static` and `binfmt-user-static` from the AUR.
2. Download and install the [Arch Linux ARM keyring](https://archlinuxarm.org/packages/any/archlinuxarm-keyring):
`pacman -U archlinuxarm-keyring-20140119-1-any.pkg.tar.xz`
3. Download Arch Linux ARM mirrorlist and pacman.conf:
https://archlinuxarm.org/packages/aarch64/pacman/files/pacman.conf
https://archlinuxarm.org/packages/any/pacman-mirrorlist/files/mirrorlist
4. Replace CARCH in pacman.conf with desired arch (aarch64 or armv7h):
`sed -i 's/@CARCH@/aarch64/' pacman.conf` or `sed -i 's/@CARCH@/armv7h/' pacman.conf`
5. Replace `/etc/pacman.d/mirorlist` with local mirrorlist:
`sed -i "s|Include = /etc/pacman.d|Include = $(pwd)|" pacman.conf`
6. Run `pacstrap` with `-G`, `-M`, and `-C` to ensure use of ARM pacman.conf and mirrorlist:
`pacstrap -GM -C pacman.conf /path/to/destination whatever packages here`

leo60228:
For what it's worth, I used these instructions as a guideline and I successfully installed using shared storage with `fdisk` instead of `parted`. I had to change `Image` to `zImage`, though.

leo60228:
> Instructions to install using `pacstrap` and a UEFI bootable USB drive, just like "normal" Arch Linux would be welcome. It would make things easier for everyone to start considering those ARM platforms that can do standards-based boots "normal" platforms and simply do the usual installation steps.

I found [multiple](https://github.com/JackMyers001/archiso-aarch64) [people](https://github.com/IComplainInComments/archiso/tree/Arch-Linux-Arm) who managed to build UEFI installer ISOs for aarch64, but I'm not sure that the Tow-Boot docs should be linking unofficial installers. Additionally, these are both only for aarch64, while these instructions work for armv7.

vincele:
FWIW, I used @benthetechguy's instruction on the pine64 wiki (which are similar to the ones proposed here) on a pinebookpro. Tested tow boot (on SD card, emmc disconnected, SPI empty) : pine64-pinebookPro-2021.10-005. This is nicely working. Thanks.

EDIT: and then borrowing on those I added a page to install void linux the same way [here](https://wiki.pine64.org/wiki/Installing_Void_Linux_ARM_On_The_Pinebook_Pro).

ALSEDD01 (approved):

PR #54, opened 2021-09-27
[WIP] kobol-helios64: init

samueldr:
~~Works, not at all ready for prime-time. I don't like how it's a huge vendor-flavoured patch.~~

~~But what's important is that **it works**.~~

~~It can safely be installed to the SPI flash.~~

~~Building for SPI Flash use is not working *at the current time*, fixing it for proprietary DDR training will be required.~~

See: https://github.com/Tow-Boot/Tow-Boot/pull/54#issuecomment-1257067670

### DIFF
--- a/boards/kobol-helios64/blobs.nix
+++ b/boards/kobol-helios64/blobs.nix
@@ -1,0 +1,61 @@
+{ runCommandNoCC, fetchurl, xxd }:
+
+#
+# Notes about serial baud rate change:
+#
+# > Find `60e3 16__`; 0x16e360, 1500000; it might be "packed" with another value.
+# > Then change to 0x1c200, 115200, `00c2 01__`
+# > Finding the offset: `grep '60 \?e3 \?16'`
+# > It's likely wou will find two. Guess which one.
+#
+
+{
+  ram_init-1_24-666 = 
+    runCommandNoCC "rk3399-patched-666-1.24-ram_init" {
+      nativeBuildInputs = [
+        xxd
+      ];
+      ram_init = fetchurl {
+        url = "https://github.com/rockchip-linux/rkbin/blob/cc30c9a0bf291d604e7ceb91ddeb5f472270a31b/bin/rk33/rk3399_ddr_666MHz_v1.24.bin?raw=true";
+        sha256 = "sha256-GlBVqL4o/u8VoFbWDl6t5ew+i7asCrh9pQafHDypRnI=";
+      };
+    } ''
+      cat $ram_init > $out
+      xxd -r - $out <<EOF
+      00014d70: 2003 0000 00c2 0123 0000 0000 0000 0103
+      EOF
+    ''
+  ;
+  ram_init-1_24-800 = 
+    runCommandNoCC "rk3399-patched-800-1.24-ram_init" {
+      nativeBuildInputs = [
+        xxd
+      ];
+      ram_init = fetchurl {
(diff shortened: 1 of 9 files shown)


lheckemann:
Tested this together with @mic92 tonight.
- eMMC doesn't seem to work, at least on his helios64; reports `Card did not respond to voltage select! : -110`, both when u-boot is installed to eMMC (SPL runs, but fails to load "real" u-boot from eMMC) and when it's loaded from the SD card and tries to scan the eMMC
- Loads fine from SD card
- Can boot from SD card
- Can boot from USB stick
- Doesn't seem to recognise mSATA SSD or SATA HDDs (might be user error?)

samueldr:
My helios64 is, annoyingly, in a "production critical" [to my home infra] role, so I can't easily test *anything*.

And more annoyingly, the fact that it's not been running stable[^1] means I need to retire it, so at some point I might be able to re-test things more :/.

Though I can state that eMMC works on my end, as it is using it to boot:

```
 $ mount | grep -i mmc
/dev/mmcblk2p1 on / type ext4 (rw,relatime)
```

To be totally frank, the memory issues and the vendor folding really discouraged me about the platform :/.

* * *

[^1]: what's not stable? Unclear. It might still be related to the underlying memory stability issues, but way harder to reproduce than the initial issues observed. The kernel crashes generally under load involving I/O.

lheckemann:
> To be totally frank, the memory issues and the vendor folding really discouraged me about the platform :/.

yeah, it's a real shame. They did so many things right.

samueldr:
While I just now updated the branch with a rebase on the current state of Tow-Boot, in addition to some other WIP rockchip changes, **I am declaring this hardware as *unfit for any uses*.**

I am closing this PR, and never supporting this board. After stress-testing the device with the lowest ram speed training available, I have reproduced the fickleness of the device *yet again*.

I suspect that anyone having any form of success with this hardware either hasn't pushed the thing far enough to fail noisily, or worse, is living with silent corruption.

I am open to reconsidering this in the future, if someone has good examples of working and stable configurations. It would need to be proven stable under *stress* conditions. Notes on how I stressed the machine:

Ran each in their own loop:

 - nix-instantiate '<nixpkgs/nixos>' -A config.system.build.toplevel
 - memtester 2G
 - `grep -R a.c.e.g.i.k.m.o.q.s.u.w.y .` on a big dir
 - `tree` on a big dir

The "big dir" is stored on a single ZFS filesystem.

I also was using it heavily through NFS during testing to hasten an expected crash. Note that *NFS has already been verified to not be the cause*, I ran the system with other configurations and ended-up with instability there too.

Since other reports online seems to corroborate my experience here, I have to say that my opinion is that the Helios64 is unfit for any use.

axelsimon:
2 years later, it appears some folks of the Armbian forum have managed to get a stable build / setup, and the device is once again a (community) supported Armbian device.

There are reports of stable Armbian Helios64 under load, and [this thread](https://forum.armbian.com/topic/30074-helios64-armbian-2308-bookworm-issues-solved/), and specifically [this post](https://forum.armbian.com/topic/30074-helios64-armbian-2308-bookworm-issues-solved/page/8/#comment-191029) provide quite a lot of info on how to get to that state.

In light of this, it may be worth revisiting!

samueldr:
(Update from the Matrix discussion)

It turns out the *recommended* blobs are the same (v1.24)

Otherwise the "current" blob used is v1.25, but the deeply linked issue has people having just as much YMMV with 1.25 as 1.24.

Issue #59, opened 2021-11-06
Raspberry Pi Zero 2 support


colemickens:
Just opening this for a place to track, share notes, etc.

samueldr:
Note that when discussing with @leo60228, I realized there is a [generic `rpi_arm64` U-Boot defconfig](https://source.denx.de/u-boot/u-boot/-/blob/master/configs/rpi_arm64_defconfig), which may be better to use, rather than smushing together two U-boot binaries in the platform firmware partition.

leo60228:
From some testing, that defconfig requires the Broadcom firmware to provide a mainline-flavored DT. This can be done with `upstream_kernel=1` in config.txt, which will look for a DTB with a mainline filename and fall back to using `overlays/upstream.dtbo` with a vendor-flavored DT (presumably `overlays/upstream-pi4.dtbo` is used on Pi 4, but isn't mentioned in docs).

boosterl:
With the [current release](https://github.com/Tow-Boot/Tow-Boot/releases/tag/release-2021.10-004) it is already possible to boot Tow-Boot on the 02W by adding a conditional filter to the `config.txt` file. Adding a conditional filter for the pi02 which selects the `Tow-Boot.noenv.rpi3.bin` kernel makes the system bootable. More information about the conditional filters can be found [here](https://www.raspberrypi.com/documentation/computers/config_txt.html#conditional-filters). The config file looks like this after the change;
```
[pi02]
kernel=Tow-Boot.noenv.rpi3.bin

[pi3]
kernel=Tow-Boot.noenv.rpi3.bin

[pi4]
kernel=Tow-Boot.noenv.rpi4.bin
enable_gic=1
armstub=armstub8-gic.bin
disable_overscan=1

[all]
arm_64bit=1
enable_uart=1
avoid_warnings=1
```
I also added the pi02W dtb file (`bcm2710-rpi-zero-2-w.dtb`), I do not know whether this is actually necessary.

leo60228:
I'd be concerned about relying on that, since it's embedding a Pi 3 device tree.

boosterl:
That's right, but I presumed it would work because I read [here](https://www.raspberrypi.com/news/new-raspberry-pi-zero-2-w-2/):

[...]

PR #61, opened 2021-11-07
raspberryPi-*: unify Pi 3/4

leo60228:
This unifies Pi 3 and Pi 4 to use a single binary, which as a side-effect makes #59 trivial. I've not tested this beyond building yet, and preferably it should be tested on every BCM2710/BCM2711 Pi board (but at least the 3B/3B+/4B).

### DIFF
--- a/boards/raspberryPi-aarch64/default.nix
+++ b/boards/raspberryPi-aarch64/default.nix
@@ -1,43 +1,25 @@
 { config, lib, pkgs, ... }:
 
 let
-  inherit (config.helpers)
-    composeConfig
-  ;
-  raspberryPi-3 = composeConfig {
-    config = {
-      device.identifier = "raspberryPi-3";
-      Tow-Boot.defconfig = "rpi_3_defconfig";
-    };
-  };
-  raspberryPi-4 = composeConfig {
-    config = {
-      device.identifier = "raspberryPi-4";
-      Tow-Boot.defconfig = "rpi_4_defconfig";
-    };
-  };
-
   configTxt = pkgs.writeText "config.txt" ''
-    [pi3]
-    kernel=Tow-Boot.noenv.rpi3.bin
-
     [pi4]
-    kernel=Tow-Boot.noenv.rpi4.bin
     enable_gic=1
     armstub=armstub8-gic.bin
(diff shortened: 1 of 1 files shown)


colemickens:
From:
 
https://patchwork.ozlabs.org/project/uboot/patch/20191106144104.28177-1-matthias.bgg@kernel.org/

> Provide a defconfig which allows us to boot Raspberrry Pi 4
and Raspberry Pi 3 Model B/B+
> Instead of using the embedded DTB as done in RPi3 we use the
devicetree provided by the firmware.

I am not sure I totally understand these implications.

Does this mean that the DTBs copied in with this PR are loaded by the rpi firmware and then provided to u-boot? What happens if they differ from the kernel?

When I built this, it booted my 5.13 extlinux entries but not the 5.14 or 5.15 entries. I'm still looking into it.

colemickens:
Also, the dtbs in my extlinux dtbs dirs for the rpi3 are named differently than the commit targets.

bcm2837-rpi-3-b.dtb rather than `bcm2710*`.

colemickens:
I guess this is copying dtbs from the rpi-firmware. Which from the patchwork thread, seems to be what happens when u-boot is embedding their own copy of the dtb anyway... :/

colemickens:
Okay, my issue was because we need to port the updated `0001-configs-rpi-allow-for-bigger-kernels.patch` in from nixpkgs. This PR seems to boot my rpi4 at least.

colemickens:
Okay, now that I've tested this with the larger kernel, and I've reviewed the notes about this defconfig, I'm in favor of removing draft status and merging. I can send a separate PR for updating the patch from nixpkgs.

(also, early testing indicates this "just works" with `rpi02`!)

samueldr:
Any particular reason for this still being a draft?

leo60228:
Oh, good point.

samueldr:
I'll test this at least in the cycle for -004, but given @colemickens' recent attempts, it seems like it's going to be an easy replacement.

leo60228:
Merge conflicts fixed. (note to self: test this more thoroughly, you have a Pi 3B+, 4B, and Zero 2 W...)

PR #63, opened 2021-11-23
Add support for Radxa RockPi4C

CRTified:
This PR adds support for the [Radxa RockPi 4C](https://wiki.radxa.com/Rockpi4/hardware). 
Due to the close relation to the other models, they are probably also easy to add.

Tested:

 - Booting from NVMe (Crucial P2 CT250P2SSD8)
 - Boot Menu (over `screen`)

Untested:

 - ~~Graphical output (Board has microHDMI and mDP, no output on mDP)~~ Output is on microHDMI

For flashing, I followed [the official instructions for flashing over USB in maskmode](https://wiki.radxa.com/Rockpi4/dev/spi-install).

### DIFF
--- a/boards/radxa/default.nix
+++ b/boards/radxa/default.nix
@@ -1,0 +1,16 @@
+{ rockchipRK399 }:
+
+{
+  radxa-RockPi4C = rockchipRK399 {
+    boardIdentifier = "radxa-rockPi4C";
+    defconfig = "rock-pi-4c-rk3399_defconfig";
+    SPISize = 4 * 1024 * 1024; # 4 MiB
+    patches = [
+      # Based on https://github.com/armbian/build/blob/master/patch/u-boot/u-boot-rockchip64/board-rock-pi-4-enable-spi-flash.patch
+      ./0001-rockpi4-rk3399-add-spi-support.patch
+      # From https://github.com/armbian/build/blob/master/patch/u-boot/u-boot-rockchip64/general-add-xtx-spi-nor-chips.patch
+      ./general-add-xtx-spi-nor-chips.patch
+    ];
+    withSPI = true;
+  };
+}
--- a/boards/radxa/0001-rockpi4-rk3399-add-spi-support.patch
+++ b/boards/radxa/0001-rockpi4-rk3399-add-spi-support.patch
@@ -1,0 +1,135 @@
+diff --git a/arch/arm/dts/rk3399-rock-pi-4-u-boot.dtsi b/arch/arm/dts/rk3399-rock-pi-4-u-boot.dtsi
+index c17e769f64..97a5be3dc1 100644
+--- a/arch/arm/dts/rk3399-rock-pi-4-u-boot.dtsi
++++ b/arch/arm/dts/rk3399-rock-pi-4-u-boot.dtsi
+@@ -8,7 +8,27 @@
+ 
+ / {
+ 	chosen {
+-		u-boot,spl-boot-order = "same-as-spl", &sdhci, &sdmmc;
++		u-boot,spl-boot-order = "same-as-spl", &spi_flash, &sdmmc, &sdhci;
++	};
++
++	config {
++		u-boot,spl-payload-offset = <0x80000>;
++	};
++};
++
++&spi1 {
++	status = "okay";
++
++	flash@0 {
++		compatible = "jedec,spi-nor";
++		reg = <0>;
++		spi-max-frequency = <10000000>;
++	};
++};
++
++&spi1 {
++	spi_flash: flash@0 {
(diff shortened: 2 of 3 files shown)


boards/radxa/0001-rockpi4-rk3399-Configure-SPI-flash-boot-offset.patch
```diff
@@ -0,0 +1,135 @@
+diff --git a/arch/arm/dts/rk3399-rock-pi-4-u-boot.dtsi b/arch/arm/dts/rk3399-rock-pi-4-u-boot.dtsi
```
samueldr: Rename to something like "Add support for SPI", as it's not strictly about the offset.

boards/radxa/0001-rockpi4-rk3399-Configure-SPI-flash-boot-offset.patch
```diff
@@ -0,0 +1,135 @@
...
+@@ -8,7 +8,27 @@
+ 
+ / {
+ 	chosen {
+-		u-boot,spl-boot-order = "same-as-spl", &sdhci, &sdmmc;
++		u-boot,spl-boot-order = "same-as-spl", &spi_flash, &sdmmc, &sdhci;
++	};
++
++	config {
++		u-boot,spl-payload-offset = <0x80000>; /* @ 384KB */
```
samueldr:
```suggestion
+		u-boot,spl-payload-offset = <0x80000>;
```

boards/radxa/default.nix
```diff
@@ -0,0 +1,16 @@
+{ rockchipRK399 }:
+
+{
+  radxa-RockPi4C = rockchipRK399 {
+    boardIdentifier = "radxa-RockPi4C";
```
samueldr:
```suggestion
    boardIdentifier = "radxa-rockPi4C";
```

The board name is `camelCased` and not `PascalCased`. Otherwise good noticing it is `Cased` :)

boards/radxa/default.nix
```diff
@@ -0,0 +1,16 @@
...
+    defconfig = "rock-pi-4c-rk3399_defconfig";
+    SPISize = 4 * 1024 * 1024; # 4 MiB
```
samueldr, line 7: > A 4MB(32Mbit) SPI flash is pasted by default, for future SPI booting support

Aww... This puts a damper on some plans I had that required *just a bit more space*... But good to know this before going forward.

(You don't need to do anything about this...)

CRTified:
Below is also a test of cycling through erasing and re-flashing using `spi.installer.img`.


---

Erasing:
```
SF: Detected XT25F32B with page size 256 Bytes, erase size 4 KiB, total 4 MiB
Currently erasing...
SF: 4194304 bytes @ 0x0 Erased: OK
Done!
```

now rebooting with the menu, starting
```
Tow-Boot 2021.04 (Apr 05 2021 - 15:03:29 +0000)003-pre [variant: noenv]
```

and now flashing it again (got dropped into that menu directly)
```
Found U-Boot script /boot.scr
-> Reading new firmware from storage...
1560164 bytes read in 90 ms (16.5 MiB/s)
-> Hardening against failures...
   We are deliberately breaking the initial part of the SPI Flash contents.
SF: 8192 bytes @ 0x0 Erased: OK
                                                                                                                                                                                                                
   A stray reboot should be safe now.
-> Writing new firmware tail to SPI Flash...
device 0 offset 0x2000, size 0x17ae64
1551972 bytes written, 0 bytes skipped in 25.500s, speed 62314 B/s
-> Writing new firmware head to SPI Flash...
device 0 offset 0x0, size 0x2000
8192 bytes written, 0 bytes skipped in 0.157s, speed 52428 B/s
✅ Flashing seems to have been successful!
Press any key to reboot...
```

pulling the cable to power it off...

Removing SD-Card

booting
```
Tow-Boot 2021.04 (Apr 05 2021 - 15:03:29 +0000)003-pre [variant: spi]
```

with bad CRC (indicating that the config got reset)

boards/radxa/default.nix
CRTified: They're easy to replace, they have a big footprint (and some were even delivered without a flash).

samueldr:
Thanks! At a glance this looks good, I'll have to run the build locally *just to verify* then it looks like it's going to be good to go. Expect to see this either in the next update, or the one following (depending if I do a "soft" release strictly to upgrade U-Boot or not).

CRTified:
Today, the microHDMI to HDMI adapter arrived and I got video output:

123f44d7-ae22-4cf6-b3b1-82c8b5d55259


So that is confirmed to work now, too :)

samueldr:
Thank you again!

PR #56, opened 2021-10-18
Tow-Boot: update U-Boot to 2021.10

samueldr:
**This still needs some testing**.

* * *

(Original description)

Only one target platform was tested, and not entirely so.

All boards built though.

I am only opening this to "show my work", and hopefully help any potential users wanting to get their hands dirty with the newer U-Boot release.

I will look in taking some time to test this correctly on all the major SoC families soon enough™ to cut the next release with only this upgrade.

* * *

### TODO

 * [x] check if poweroff now works on RK3399
 * [x] Check ^C behaviour at boot
 * [x] Add workaround for eMMC and rk3399 (see PPP PR)

 - [x] Test on hardware

PR #65, opened 2021-12-08
Add pine h64 board

dwagenk:
This adds support for the pine H64 board, both the A and the B variant (which are almost identical). An adapted builder for the Arm-Trusted-Firmware for the Allwinner H6 SOC was also necessary.

https://wiki.pine64.org/index.php/PINE_H64

I tested the `shared.disk-image.img`  with the B variant of the board.

--- List updated ---
- [x] Tow-Boot is starting and I get the boot selection menu via serial connection when pressing CTRL-C or ESC.
- [x] Switching out the SD Card and selecting "Boot from SD" works
- [x] USB: lower of the two USB2 port works
- [ ] USB: The upper USB2 port one and the USB3 Port do not work (even after update to v2021.10)
- [ ] HDMI: not working (apparently no support in u-boot, see `arch/arm/mach-sunxi/Kconfig` dependecy chain `VIDEO_HDMI->VIDEO_SUNXI->!SUN50I_GEN_H6`)
- [x] netbooting (DHCP or PXE): at least the IP gets assigned, did *not* set up a TFTP server for testing
- [ ] eMMC: I do not have an eMMC module and thus didn't test booting form it
- [x] SPI flash: works, though BOOTROM attempts to boot from SD and eMMC first

### DIFF
--- a/support/overlay/overlay.nix
+++ b/support/overlay/overlay.nix
@@ -54,6 +54,7 @@ in
 
     inherit (callPackage ./arm-trusted-firmware { })
       armTrustedFirmwareAllwinner
+      armTrustedFirmwareAllwinnerH6
       armTrustedFirmwareRK3399
       armTrustedFirmwareS905
     ;
--- a/boards/pine64-pineH64A/0001-sunxi-dts-add-device-tree-for-pine-H64-model-B.patch
+++ b/boards/pine64-pineH64A/0001-sunxi-dts-add-device-tree-for-pine-H64-model-B.patch
@@ -1,0 +1,90 @@
+From 40b6a302e4e1613476e6a92620d7ceefb239831f Mon Sep 17 00:00:00 2001
+From: Daniel Wagenknecht <dwagenk@mailbox.org>
+Date: Sun, 12 Dec 2021 01:15:54 +0100
+Subject: [PATCH 1/4] sunxi: dts: add device-tree for pine H64 model B
+
+Add the device-tree for pine H64 model B from 5.15 mainline linux.
+
+The differences between the pine H64 boards model A and model B are
+irrelevant inside u-boot, but the device-tree name is relevant as
+default for selecting the correct device-tree when booting linux.
+
+Signed-off-by: Daniel Wagenknecht <dwagenk@mailbox.org>
+---
+ arch/arm/dts/Makefile                       |  1 +
+ arch/arm/dts/sun50i-h6-pine-h64-model-b.dts | 51 +++++++++++++++++++++
+ 2 files changed, 52 insertions(+)
+ create mode 100644 arch/arm/dts/sun50i-h6-pine-h64-model-b.dts
+
+diff --git a/arch/arm/dts/Makefile b/arch/arm/dts/Makefile
+index fc16a57e60..de2e4ca66b 100644
+--- a/arch/arm/dts/Makefile
++++ b/arch/arm/dts/Makefile
+@@ -642,6 +642,7 @@ dtb-$(CONFIG_MACH_SUN50I_H6) += \
+ 	sun50i-h6-orangepi-lite2.dtb \
+ 	sun50i-h6-orangepi-one-plus.dtb \
+ 	sun50i-h6-pine-h64.dtb \
++	sun50i-h6-pine-h64-model-b.dtb \
+ 	sun50i-h6-tanix-tx6.dtb
+ dtb-$(CONFIG_MACH_SUN50I_H616) += \
+ 	sun50i-h616-orangepi-zero2.dtb
+diff --git a/arch/arm/dts/sun50i-h6-pine-h64-model-b.dts b/arch/arm/dts/sun50i-h6-pine-h64-model-b.dts
+new file mode 100644
+index 0000000000..686f58e770
+--- /dev/null
++++ b/arch/arm/dts/sun50i-h6-pine-h64-model-b.dts
+@@ -0,0 +1,51 @@
++// SPDX-License-Identifier: (GPL-2.0+ or MIT)
++/*
++ * Copyright (C) 2019 Corentin LABBE <clabbe@baylibre.com>
++ */
++
++#include "sun50i-h6-pine-h64.dts"
++
++/ {
++	model = "Pine H64 model B";
++	compatible = "pine64,pine-h64-model-b", "allwinner,sun50i-h6";
++
++	/delete-node/ reg_gmac_3v3;
++
++	wifi_pwrseq: wifi_pwrseq {
++		compatible = "mmc-pwrseq-simple";
++		reset-gpios = <&r_pio 1 3 GPIO_ACTIVE_LOW>; /* PM3 */
++		post-power-on-delay-ms = <200>;
++	};
++};
++
++&hdmi_connector {
++	/delete-property/ ddc-en-gpios;
++};
++
++&emac {
++	phy-supply = <&reg_aldo2>;
++};
++
++&mmc1 {
++	vmmc-supply = <&reg_cldo3>;
++	vqmmc-supply = <&reg_aldo1>;
++	mmc-pwrseq = <&wifi_pwrseq>;
++	bus-width = <4>;
++	non-removable;
++	status = "okay";
++};
++
++&uart1 {
++	pinctrl-names = "default";
++	pinctrl-0 = <&uart1_pins>, <&uart1_rts_cts_pins>;
++	uart-has-rtscts;
++	status = "okay";
++
++	bluetooth {
++		compatible = "realtek,rtl8723bs-bt";
++		device-wake-gpios = <&r_pio 1 2 GPIO_ACTIVE_HIGH>; /* PM2 */
++		host-wake-gpios = <&r_pio 1 1 GPIO_ACTIVE_HIGH>; /* PM1 */
++		enable-gpios = <&r_pio 1 4 GPIO_ACTIVE_HIGH>; /* PM4 */
++		max-speed = <1500000>;
++	};
++};
+-- 
+2.33.1
+
--- a/boards/pine64-pineH64A/0002-sunxi-configs-add-defconfig-for-pine-H64-model-B.patch
+++ b/boards/pine64-pineH64A/0002-sunxi-configs-add-defconfig-for-pine-H64-model-B.patch
@@ -1,0 +1,48 @@
+From df5c906043659276240e62ce003217ecea545bdb Mon Sep 17 00:00:00 2001
+From: Daniel Wagenknecht <dwagenk@mailbox.org>
+Date: Wed, 8 Dec 2021 23:05:42 +0100
+Subject: [PATCH 2/4] sunxi: configs: add defconfig for pine H64 model B
+
+It almost identical to the defconfig for pine H64 model A with only
+the CONFIG_DEFUALT_DEVICE_TREE adapted.
+
+The differences between the boards model A and model B are irrelevant
+inside u-boot, but the device-tree name is relevant as default for
+selecting the correct device-tree when booting linux.
+
+Signed-off-by: Daniel Wagenknecht <dwagenk@mailbox.org>
+---
+ configs/pine_h64-model-b_defconfig | 21 +++++++++++++++++++++
+ 1 file changed, 21 insertions(+)
+ create mode 100644 configs/pine_h64-model-b_defconfig
+
+diff --git a/configs/pine_h64-model-b_defconfig b/configs/pine_h64-model-b_defconfig
+new file mode 100644
+index 0000000000..4645ceaa54
+--- /dev/null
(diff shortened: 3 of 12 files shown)


support/overlay/overlay.nix
```diff
@@ -78,6 +79,10 @@ in
       TF-A = aarch64.armTrustedFirmwareAllwinner;
     };
 
+    allwinnerH6 = aarch64.callPackage ../builders/allwinner-a64 {
+      TF-A = aarch64.armTrustedFirmwareAllwinnerH6;
+    };
+
```
samueldr: (You don't need to do anything about this...)

I'm rewriting the build setup for Tow-Boot using the modules system. So *importing the A64 builder for H6* won't be a thing anymore in the future. So it's fine to add this right now even though I don't exactly like this :).

Using the modules system, it will be easier to configure complex build instructions.

boards/pine64/0002-configs-add-defconfig-for-pine-H64-Model-B.patch
```diff
@@ -0,0 +1,39 @@
...
++CONFIG_MMC0_CD_PIN="PF6"
++CONFIG_MMC_SUNXI_SLOT_EXTRA=2
++CONFIG_USB3_VBUS_PIN="PL5"
++# CONFIG_PSCI_RESET is not set
++# CONFIG_SYS_MALLOC_CLEAR_ON_INIT is not set
++CONFIG_USE_PREBOOT=y
++# CONFIG_CMD_FLASH is not set
++# CONFIG_SPL_DOS_PARTITION is not set
++# CONFIG_SPL_EFI_PARTITION is not set
++CONFIG_DEFAULT_DEVICE_TREE="sun50i-h6-pine-h64-model-b"
```
samueldr: (You don't need to do anything about this, this is a not for future readers wanting to know more about porting)

It is important that `CONFIG_DEFAULT_DEVICE_TREE` matches the file names in the mainline kernel.

 - https://github.com/torvalds/linux/blob/2a987e65025e2b79c6d453b78cb5985ac6e5eb26/arch/arm64/boot/dts/allwinner/sun50i-h6-pine-h64.dts
 - https://github.com/torvalds/linux/blob/2a987e65025e2b79c6d453b78cb5985ac6e5eb26/arch/arm64/boot/dts/allwinner/sun50i-h6-pine-h64-model-b.dts

This is because this config is used to seed the environment with the name of the `dtb` file that should be loaded automatically with the different boot schemes.

boards/pine64/default.nix
```diff
@@ -14,6 +14,20 @@
       ./0001-configs-pine64-lts-Enable-SPI-flash.patch
     ];
   };
+  pine64-pineH64A = allwinnerH6 {
+    boardIdentifier = "pine64-pineH64A";
+    defconfig = "pine_h64_defconfig";
+    withPoweroff = false;
```
samueldr: Why `withPoweroff = false;`? (Not that I doubt it could be non-working.)

samueldr:
Hi! :wave: A new face! And welcome.

Looks good generally speaking.

Are you going to look at SPI Flash for this PR?

Other non-working features (e.g. HDMI) are more for upstream problems, assuming it's not just not configured. We need to upgrade the backing U-Boot version anyway, so it's not too much of a concern for now. It will need to be re-evaluated once udpated.

dwagenk:
I'm not sure what else needs to be done for enabling SPI. I'm planning to dive in a little deeper to see if I can fix this. The trivial approach of enabling the option in the Tow-Boot board description and patching the `defconfig` to include
`
CONFIG_SPL_SPI_SUNXI=y
CONFIG_SPI_FLASH_WINBOND=y
`
didn't work. I think there's some intermediary Kconfig option missing and thus the config options from the `defconfig` are not properly applied. First found problem is 
```
Define one of CONFIG_ENV_IS_IN_{EEPROM|FLASH|MMC|FAT|EXT4|NAND|NVRAM|ONENAND|SATA|SPI_FLASH|REMOTE|UBI} or CONFIG_ENV_IS_NOWHERE
```
while that is explicitely set through the Tow-Boot builder.

The `.dts` file mentions eMMC and SPI sharing a pin, so eMMC would probably have to be disabled in there.

> Are you going to look at SPI Flash for this PR?
I'd prefer to have this merged without SPI and then do a follow-up PR if I get SPI to work.

boards/pine64/default.nix
dwagenk: `withPoweroff = false;` seemed to be set for most SBCs in Tow-Boot. I'll enable it and see if the poweroff command works as expected.

samueldr: Hopefully it will work. It works for existing supported Allwinner boards.

samueldr:
Ooooof... In other words, how to make the SPI basically unusable :|

 - https://source.denx.de/u-boot/u-boot/-/blob/v2021.10/arch/arm/dts/sun50i-h6-pine-h64.dts#L301-317

AFAIUI Tow-Boot could be made to load from SPI, but that build wouldn't be able to interact with SPI once loaded. E.g. it wouldn't be able to save env, or update itself. I'm not sure this use case is already "supported" in U-Boot. E.g. SPL has `&mmc2` disabled, and `&spi0` enabled, but the "proper" U-Boot itself has `&mmc2` enabled and `&spi0` disabled.

I believe your issues here with enabling SPI were that it wasn't configured in the device tree. But that's a guess.

Though it might be possible to add some sort of special casing to "unprobe" the MMC2 while reading/writing to SPI. This would require out of scope work, especially for this PR.

dwagenk:
> AFAIUI Tow-Boot could be made to load from SPI, but that build wouldn't be able to interact with SPI once loaded. E.g. it wouldn't be able to save env, or update itself. I'm not sure this use case is already "supported" in U-Boot. E.g. SPL has `&mmc2` disabled, and `&spi0` enabled, but the "proper" U-Boot itself has `&mmc2` enabled and `&spi0` disabled.

Is it possible to reload the dts from inside u-boot? I'm thinking about something like
1. Boot with `&mmc2` disabled and `&spi0` enabled
2. Only when "Boot from eMMC" is selected from the menu switch to `&mmc2` enabled and `&spi0` disabled, probe mmc2

> I believe your issues here with enabling SPI were that it wasn't configured in the device tree. But that's a guess.

I don't think so. I'm running into build problems with SPI, but to my understanding u-boot sees device-tree as run-time config (like linux, unlike zephyr) and doesn't pull any data from it during build.

boards/pine64/default.nix
dwagenk: `withPoweroff = true` gives build errors.

```
builder for '/nix/store/4cs196kifj0dmmddacsp5qy12ynqfmdq-tow-boot-pine_h64-model-b_defconfig-noenv-aarch64-unknown-linux-gnu-2021.10-003-pre.drv' failed with exit code 1; last 10 log lines:
    AR      net/built-in.o
    LDS     u-boot.lds
    LD      u-boot
  /nix/store/hkwgs14jbngmxs5z6v918szwpcs474xb-aarch64-unknown-linux-gnu-binutils-2.35.1/bin/aarch64-unknown-linux-gnu-ld.bfd: lib/efi_loader/efi_runtime.o: in function `efi_reset_system_boot
time':
  /build/u-boot-2021.10/lib/efi_loader/efi_runtime.c:217: undefined reference to `do_poweroff'
  /nix/store/hkwgs14jbngmxs5z6v918szwpcs474xb-aarch64-unknown-linux-gnu-binutils-2.35.1/bin/aarch64-unknown-linux-gnu-ld.bfd: cmd/boot.o:(.u_boot_list_2_cmd_2_poweroff+0x18): undefined refer
ence to `do_poweroff'
  /nix/store/fmngyhzl5rfz23xnc7lrgkc95mqzx8hj-aarch64-unknown-linux-gnu-stage-final-gcc-debug-wrapper-9.3.0/bin/aarch64-unknown-linux-gnu-ld.bfd: line 249: 10688 Segmentation fault      /nix
/store/hkwgs14jbngmxs5z6v918szwpcs474xb-aarch64-unknown-linux-gnu-binutils-2.35.1/bin/aarch64-unknown-linux-gnu-ld.bfd ${extraBefore+"${extraBefore[@]}"} ${params+"${params[@]}"} ${extraAfte
r+"${extraAfter[@]}"}
  make: *** [Makefile:1790: u-boot] Error 139
  make: *** Deleting file 'u-boot'
  builder for '/nix/store/4cs196kifj0dmmddacsp5qy12ynqfmdq-tow-boot-pine_h64-model-b_defconfig-noenv-aarch64-unknown-linux-gnu-2021.10-003-pre.drv' failed with exit code 2
cannot build derivation '/nix/store/jcmx13ck4wk0423a8ancd5y6fy0ql1gl-tow-boot-pine_h64-model-b_defconfig-noenv-aarch64-unknown-linux-gnu-2021.10-003-pre.drv': 1 dependencies couldn't be buil
t
```

dwagenk:
- rebased onto https://github.com/Tow-Boot/Tow-Boot/pull/56
- enabled & fixed SPI
- reworked the patches a little

Patches `0001 - 0003` will be sent to the u-boot mailing list and thus might be obsolete soon.

I updated the PR description to give an overview. I'm currently not planning on doing anything about the missing HDMI support and the non working USB ports.

Regarding the eMMC I can't test without a module. Since u-boot doesn't do mutlithreading I guess enabling both eMMC and SPI might actually work. If the drivers ensure the pinmuxing is configured before operating on the busses they should both work.

boards/pine64/0002-sunxi-configs-add-defconfig-for-pine-H64-model-B.patch
```diff
@@ -0,0 +1,48 @@
+From df5c906043659276240e62ce003217ecea545bdb Mon Sep 17 00:00:00 2001
+From: Daniel Wagenknecht <dwagenk@mailbox.org>
+Date: Wed, 8 Dec 2021 23:05:42 +0100
+Subject: [PATCH 2/4] sunxi: configs: add defconfig for pine H64 model B
+
+It almost identical to the defconfig for pine H64 model A with only
+the CONFIG_DEFUALT_DEVICE_TREE adapted.
```
samueldr:
```suggestion
the CONFIG_DEFAULT_DEVICE_TREE adapted.
```

(Since you want to upstream, let's correct nits :))

boards/pine64/0002-sunxi-configs-add-defconfig-for-pine-H64-model-B.patch
```diff
@@ -0,0 +1,48 @@
...
+
+The differences between the boards model A and model B are irrelevant
+inside u-boot, but the device-tree name is relevant as default for
+selecting the correct device-tree when booting linux.
```
samueldr: (Nothing to do really)

I believe the "more correct" way to handle this would be like for the Pinephone, figure out a detectable difference between boards during early boot and setup accordingly.

I don't really know the H64. There may be nothing that can be detected for this. At least from the Linux dts, I don't really see anything.

auchter:
Hey y'all,

As this branch no longer cleanly applied to the HEAD of the development branch (and I was wanting Tow-Boot on my Pine H64), I took a stab at updating this. I have a branch at https://github.com/auchter/Tow-Boot/commits/auchter/pineh64 that should apply cleanly.

The branch incorporates @dwagenk's patches against u-boot, but I rewrote the Nix stuff to deal with the changes in Tow-Boot since this PR was opened. I also cherry-picked two of @samueldr's commits into that branch from #155 to cleanly handle the lack of display support for the H64 in upstream u-boot.

I chose to add a new board for each variant (H64A/H64B) with the H64B importing and overriding the configuration as appropriate. 

Like @dwagenk, I don't have an eMMC module to test with (nor do I have an H64A), but I was able to successfully boot both `shared.disk-image.img` and `spi.installer.img` via SD on my H64B, and also verified that flashing and then booting from SPI works.

Let me know if you'd like me to open up a separate PR for this

samueldr:
I personally wouldn't mind another PR, but let's also see if @dwagenk has anything to say, just in case, maybe you can cooperate for bringing the changes in.

dwagenk:
Hey. Don't have time to continue this for now. @auchter I added you as collaborator to my fork, so you should be able to modify the base-branch for this PR.

auchter:
Thanks, @dwagenk! I've pushed my changes to the base branch, and also pushed a branch (dwagenk/add-pineH64-board-orig) to backup your original branch state for posterity.

Looks like your pinmux fix landed in u-boot already, and someone else has added a Model B device tree, so we should be able to drop patches 1 and 3 once Tow-Boot updates the u-boot version used. I'll ping the mailing list this week to see if we can get the defconfig for model B merged as well.

clee:
Just wanted to comment here, I tried @auchter's branch and was able to successfully build and install tow-boot to the SPI flash on my H64B here. And now my H64B can PXE boot without any SD card installed :tada:

Issue #47, opened 2021-06-30
doc: Document neutering other platform firmware installs


samueldr:
It might be helpful for people with existing installs that wants to install to another storage. E.g. a Pinebook Pro user that wants to remove the eMMC U-Boot from their distro from the firmware boot flow, so that SD card boot is usable.

* * *

# SPI

Erasing [the first `0x2000` bytes](https://github.com/Tow-Boot/Tow-Boot/blob/00f2cce0e270f87e8f013e476ef5db8e0f79e3b7/support/builders/spi-installer/default.nix#L79) was verified to be enough on Allwinner A64, RK3399 and S805X (and assumed to be fine for other SoCs in their respective families).

# eMMC / SD

## RK3399

Replace `/dev/mmcblkX` by the block storage you want to neuter the install on. Note that running on any other storage **can break the data it contains**. On RK3399 platforms, **generally** (please verify on your device), the eMMC is at `/dev/mmcblk2`.

```
# dd if=/dev/zero of=/dev/mmcblkX bs=512 seek=64 count=16
```

PR #67, opened 2021-12-20
Phone and tablet UX + mmcboot support

samueldr:
* * *

### *Give me the build*

Woah there, you probably want to read and review before, but eh, if you want it, here it is:

#### Build 004.5

**Built against e65e42de0817bf3c4baf9ef73d2bc33db5911d30**

| Build | Installation notes | Troubleshooting guide |
|---|---|---|
| [build 004.5 Pinephone A64 build](https://stuff.samueldr.com/7pnblffpscrmx51vlbl1vky53rizhxqv-Tow-Boot.release.004-pre/pine64-pinephoneA64-2021.10-004-pre.tar.xz) | [Installation notes](https://github.com/Tow-Boot/Tow-Boot/blob/e65e42de0817bf3c4baf9ef73d2bc33db5911d30/boards/pine64-pinephoneA64/INSTALLING.md) | *(none)* |
| [build 004.5 Pinephone Pro build](https://stuff.samueldr.com/7pnblffpscrmx51vlbl1vky53rizhxqv-Tow-Boot.release.004-pre/pine64-pinephonePro-2021.10-004-pre.tar.xz) | [Installation notes](https://github.com/Tow-Boot/Tow-Boot/blob/e65e42de0817bf3c4baf9ef73d2bc33db5911d30/boards/pine64-pinephonePro/INSTALLING.md) | [Troubleshooting guide](https://gist.github.com/samueldr/96171ff1d85b682b067753b1cf57f9e1) |

<details><summary><b>Outdated builds</b></summary>

 - ~~[WIP 1 Pinephone Pro build](https://stuff.samueldr.com/Tow-Boot/pine64-pinephonePro-2021.10-003-pre.tar.xz)~~
 - ~~[WIP 2 Pinephone Pro build](http://stuff.samueldr.com/8rqqjxr95ck6n9yaysw1r2jqpbzi3nbk-pine64-pinephonePro-2021.10-004-pre.tar.xz)~~

 - ~~Built against db86fbfd341e7f5366217bea91cfb01a948ce54a~~
     - ~~[WIP 3 Pinephone A64 build](http://stuff.samueldr.com/h4d0a6sj7jr425vpgdprk6d7bab8cy72-pine64-pinephoneA64-2021.10-004-pre.tar.xz)~~
     - ~~[WIP 3 Pinephone Pro build](http://stuff.samueldr.com/j8ylph2377b4z3rp0m4j879h7lp6l3r3-pine64-pinephonePro-2021.10-004-pre.tar.xz)~~

 - ~~Built against ccf5627f5977ebf9637d6ac495c63690f2145223; main difference: mmcboot installer for Pinephone A64.~~
     - ~~[WIP 4 Pinephone A64 build](http://stuff.samueldr.com/fdb270zzddxh41a7dyv8vgck6f3svmva-pine64-pinephoneA64-2021.10-004-pre.tar.xz)~~
     - ~~[WIP 4 Pinephone Pro build](http://stuff.samueldr.com/pfyi82x1x7zckxqbywpf31p5l49pfiij-pine64-pinephonePro-2021.10-004-pre.tar.xz)~~

</details>

* * *

### TODO

 - [x] Final reviews / approval by peers

* * *

### What is this?

This is my attempt at making boot on "Linux phones" as painless as possible.

Features:

 - LED management
 - Phone keys based control
 - User-friendlier universal defaults
 - "Status reporting"

Anti-features:

 - Superficial status reporting without display enablement.

#### LED management

The LED is lit-up as early as possible, assuming it's RGB, it's lit-up red first.

When Tow-Boot starts running the default distro boot (as defined by U-Boot), it turns yellow.

Some additional boot modes will turn the LED different colours:

 - Blue for Mass Storage
 - Aqua for *requesting* SD card boot

#### Phone keys based control

As commonly done with "consumer" phones and mobile devices, holding volume keys during early boot phases will affect the boot in different manner.

 - Volume up: Boot in USB Mass Storage mode
 - Volume down: Request Operating System Boot from SD card

> **Note**: *USB Mass Storage* (or UMS) exports the eMMC (internal) of the device over USB.

#### User-friendlier universal defaults

To make things less confusing across devices, **all** devices will load the operating system using the same rules.

The schemes supported by the distro boot scheme from U-Boot are all tried on the following devices in order:

 - eMMC
 - SD

The order is static and *eMMC first* **on all devices**. This is a deviation from the defaults from the different communities, and somewhat a deviation from the defaults of U-Boot. Doing so helps reduce confusion and should help reduce support load.

This helps reduce confusion, because the order the operating systems will be tried to be loaded is the same, wherever Tow-Boot was started from. This is quite important considering different SoCs have a different BootROM startup order, making it harder to "just" boot the SD card if started from the SD card.

Furthermore, attempting Operating System boot from eMMC first is a done considering *booting from a removable storage should be a conscious choice, **requested** by the end-user*.

> **Note**: When all Operating System boot sources have been attempted, and none worked, the device will poweroff after reporting the failure.

Not specific to the Phone UX, Tow-Boot *always uses 115200 baud rate*. This way, the UART rate is always the same, helping a bit more with unified instructions.

#### "Status reporting"

In addition to the LEDs reporting the current boot state, vibration support has been added. The device will vibrate once as early as possible.

When failing to do the Operating System boot from SD *as requested by the user* (holding volume down), the LED will flash 4 times, and the device will vibrate accordingly. Boot will continue to the default distro boot.

When failing to do the default distro boot, the LED will flash 10 times, and the device will vibrate accordingly. Then the device will poweroff.

> **Note**: when connected to a power source, some devices (the Pinephone Pro) will not poweroff, and will instead reboot.

In addition to that, with the shared disk image, a *protective partition* is added to the GPT to prevent bad manipulations from breaking the installed *Platform Firmware*. This protective partition also provides strong guarantees when updating the Tow-Boot install, such that when its size increases, it is assured not to interfere with the other partitions.

#### Superficial status reporting without display enablement

As described previously, all of this is done such that the user knows what is happening, even though the display shows nothing.

This, to me is a sort of "anti-feature", but in reality we must keep this scheme such that further devices can use it even if a display driver has not been written yet for them.

* * *

### How do I install this?

Ideally this would be installed to SPI flash using an SPI flash installer. An implementation was written with a "specialized" Linux system build so that the initial lack of display enablement on the Pinephone Pro wouldn't be an issue.

Alas, none of the targeted devices support SPI boot, thus none can be installed to dedicated storage.

To install to eMMC, the steps will generally be the same:

 - **These instructions will clobber the content of the SD card and the eMMC, defining a new partition table.**
 - Write the `shared.disk-image.img` to an SD card as-is
 - Boot the device to USB Mass Storage by holding Volume Up
 - Write the `shared.disk-image.img` to the phone.

> **Note**: With some SoCs (e.g. Rockchip) you will need to ["neuter" the *Platform Firmware*](https://github.com/Tow-Boot/Tow-Boot/issues/47) installed to the eMMC if there is one already installed. 

### How do I install a Linux distro after installing Tow-Boot?

This is a hard question to answer correctly. This will require a change in habits of the different distributions.

<details><summary>Misc. distro notes</summary>

#### Mobile NixOS

I can only answer for the distribution I'm involved with, Mobile NixOS.

A temporary installation method has been added:

 - https://github.com/NixOS/mobile-nixos/pull/442
 - https://github.com/NixOS/mobile-nixos/pull/445

It relies on a script and Mass Storage support on the target device. This is a method I think would be acceptable to keep recommending for other distros. The main thing to understand here is the script manipulates the partitions, then splats the content in them.

> The permanent solution will be an *installer system*. The user will write Tow-Boot to the eMMC, and boot from a *universal* SD card that holds an installer system that will guide the user through some questions about partitioning (e.g. ask about FDE and filesystems), the preferred graphical environment, and then do a "classic" opinionated `nixos-install` under the covers.

</details>

* * *

Handles part of #104. (Devices listing on the website.)


Danct12:
For mass storage mode, is it possible to export the SD card as well? It would be nice if you can write to SD card from there.

samueldr:
Unless I missed something when I looked, the UMS driver for U-Boot *at this point in time* only handles exporting one LUN.

With UX improvements (mainly display output support) we could let the user pick which storage medium is exported.

An alternative could be to look into adding support for exporting more LUNs in upstream U-Boot.

Be-ing:
> An alternative could be to look into adding support for exporting more LUNs in upstream U-Boot.

I think letting the user choose which device is exported would be more foolproof. With both the SD and eMMC exported, the user could accidentally write to the wrong device. With Jump Drive on the PinePhone, I've had to check the drive size reported in dmesg to ensure I'm writing to the correct device. In the event the user is using an SD card that happens to be the same size as the eMMC drive, that could be tricky.

samueldr:
This push rebases the work that was on this branch, with the refreshed work built on top of #78.

That rewrite was done **because** this PR showed it was really needed (in addition to making other things easier).

This push also includes the "ELO" changes, "Embedded Linux OS", which is used to build a touch-based installer system for installing to SPI, for devices that support installation to SPI.

Strit:
Just writing to confirm that the WIP2 builds work fine on my Pinephone Pro.
It even makes the phone able to suspend!

Does the WIP3 build add something or is it just re-arranging some stuff?

samueldr:
Re-arranging stuff, *adding a Pinephone A64 pre-built artifact*, and fixes the OS boot order regression.

But otherwise mostly the same.

boards/pine64-pinephoneA64/default.nix
```diff
@@ -0,0 +1,56 @@
...
+      (helpers: with helpers; {
+        BUTTON_GPIO = yes;
+        BUTTON_SUN4I_LRADC = yes;
+        LED_GPIO = yes;
+        VIBRATOR_GPIO = yes;
+      })
+      (helpers: with helpers; {
+        USB_MUSB_GADGET = yes;
+        USB_GADGET_MANUFACTURER = freeform ''"Pine64"'';
+      })
```
a-wai, line 45:
```suggestion
      })
      (helpers: with helpers; {
        SUPPORT_EMMC_BOOT = yes;
      })
```

This option would allow u-boot to boot from eMMC boot partitions on the A64.

a-wai: Also, upstream u-boot has support for startup LED (since 2022.01 IIRC) in case you hadn't noticed, so an additional patch isn't needed for that (vibrator isn't supported upstream, however).

(I have to admit I didn't look up what u-boot version this PR was based on though, so this comment might or might not be relevant)

samueldr: The patch implements LED startup like other RK3399 boards implement it.

I seem to remember looking at implementing it "correctly" through their framework, and that it could not be used during SPL since it brings in too much code to fit.

samueldr: > This option would allow u-boot to boot from eMMC boot partitions on the A64.

Stay tuned for the A64, work is in progress for support for eMMC boot partitions.

jussihi:
Hello, the download links don't seem to work for me for some reason. Is there some problems with the domain name redirection?

samueldr:
AFAICT it still is working just fine here. `wget` or `curl` output might help track down issues.

jussihi:
> AFAICT it still is working just fine here. `wget` or `curl` output might help track down issues.

Thank you, it actually worked, it was just a bug or something in my web browser. I was able to download by copying the link to another tab (?!). Thanks for the work, now I'm rocking Tow-Boot on my PPP! It works great!

justuser-ppp:
Hello all. I erased spi and flashed it. Now I cannot boot from sdcard. Is there any way to recover from this? Thank you for your help.

justuser-ppp:
Apparently I still can boot from sdcard if i continue spamming volume down button on boot. Everything works as it should after. Strange.

PureTryOut:
Note that you don't need to erase the SPI before flashing it, the flash button already handles that. Also please don't ask for user support here but join [#tow-boot:matrix.org](https://matrix.to/#/#tow-boot:matrix.org) instead.

justuser-ppp:
Thank you. Joined;)

samueldr:
It was beginning to be too cumbersome to separate the changes for mmcboot support, when this PR touched the installer config.

I hate bundling more features into a huge PR, but this had to be done.

chookity-pokk:
> Note that you don't need to erase the SPI before flashing it, the flash button already handles that. Also please don't ask for user support here but join [#tow-boot:matrix.org](https://matrix.to/#/#tow-boot:matrix.org) instead.

The tow-boot matrix doesn't seem to be join-able anymore?

samueldr:
That's likely because to discover rooms, sometimes "via" servers are needed, try:

 - https://matrix.to/#/#Tow-Boot:matrix.org?via=matrix.org

boards/pine64-pinephoneA64/default.nix
samueldr: Oh, forgot to come back to this discussion and say it: it's done. It works. Tested against Allwinner and Amlogic targets.

boards/pine64-pinebookA64/default.nix
```diff
@@ -1,12 +1,13 @@
 {
   device = {
     manufacturer = "PINE64";
-    name = "A64-LTS";
+    name = "Pinebook (A64)";
```
samueldr, line 4: shhhh

boards/pine64-pinephoneA64/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Additional features
+
+The phone can be started in *USB Mass Storage* mode by holding the *volume up*
+button at startup before and during the second vibration. The LED will turn
+blue if done successfully. In this mode, the phone will work like a USB drive
+when connected to a host computer.
+
+Booting an operating system, or an installer written to an SD card will require
+holding the *volume down* button before and during the second vibration. When
+done successfully, the LED will turn aqua.
```
ItsHarper:
```suggestion
Booting from an SD card will require holding the *volume down* button before and
during the second vibration. When done successfully, the LED will turn aqua.
```
The way you had it worded could be interpreted as saying "Booting an operating system will require holding down the volume down button".

boards/pine64-pinephonePro/README.md
```diff
@@ -0,0 +1,16 @@
...
+## Additional features
+
+The phone can be started in *USB Mass Storage* mode by holding the *volume up*
+button at startup before and during the second vibration. The LED will turn
+blue if done successfully. In this mode, the phone will work like a USB drive
+when connected to a host computer.
+
+Booting an operating system, or an installer written to an SD card will require
+holding the *volume down* button before and during the second vibration. When
+done successfully, the LED will turn aqua.
```
ItsHarper:
```suggestion
Booting from an SD card will require holding the *volume down* button before and
during the second vibration. When done successfully, the LED will turn aqua.
```

ItsHarper (changes requested):

ItsHarper:
I just installed this onto my A64 Pinephone. The installation process was very smooth, and thanks to the new mmcboot support, it didn't wipe my existing OS installation! Awesome work.

ItsHarper:
This PR should get linked to #76

boards/pine64-pinephoneA64/README.md
samueldr: I edited the wording to include *operating system* still, as I try very hard in the documentation to strongly delineate between *platform firmware startup* and *operating system boot*. This is because the former is generally static and set in hardware, while the latter is handled by the platform firmware, and can be anything.

PR #78, opened 2022-01-24
Rewrite eval using the modules system

samueldr:
This will allow better configuring differences between boards and hardware. Additionally, making *feature sets*, like the [phone UX](https://github.com/Tow-Boot/Tow-Boot/pull/67), will be possible without having to inject arguments everywhere.

* * *

### Notes

Most build outputs are reproducible, and verified to be equivalent.

Some will not be, because some oversights in the previous tooling were caught. Though by re-creating the oversights, the builds end up reproduced.

Only `GXL` boards were identified as non-reproducible, for the time being. The actual U-Boot build is reproducible, what isn't are the `gxlimg` manipulations. 

* * *

### TODO

 - [x] Evaluate an equivalent build through the *modules system*.
 - [x] Port all boards
 - [x] Include disk image creation back
 - [x] Produce SPI installer
 - [x] Produce main outputs (unsuitable WIP locally)
 - [x] Handle Raspberry Pi default output (ugh)
 - [x] Rewrite `release.nix`

### Can't test

 - odroid-C2
 - amlogic on shared media

Given the binaries are reproducibly the same, if something fails on amlogic, it's going to be limited to the shared media image, and not an SPI update.


samueldr:
Reproducibility changes notwithstanding, this produces builds mostly-strictly equivalent with the current state.

The main difference is how the "GPL compliance" is handled.

 - before: dump of patches, some changes might be missing
 - after: dump of the whole source tree

This dump is big, but ensures the **entire** source code is always linked to a release artifact. The history is not present, but the history is always available through this repository. By picking up the source archive, you can always diff with the upstream version to get an actual overview of the changes.

PR #83, opened 2022-01-28
support/docker: Add build shim

### LINKED ISSUE #5: Development environment for contributors (without Nix)
This project uses Nix to coordinate builds.

It's great and all, but contributors may not want, or may be unable to install Nix.

Provide a pre-configured Docker-based environment to transparently use Nix to build this.

samueldr:
This adds a shim allowing end-users to invoke a containerized Nix
instance, in a manner compatible to using `nix-build`.

```
 $ support/docker/build.sh -A pine64-pineA64LTS
```

This is equivalent to

```
 $ nix-build -A pine64-pineA64LTS
```

The semantics vary a little. *Only* the `result` symlink is resolved and
replaced with the store path contents. This means that using
`--out-link` is not supported at this time, and multiple build outputs
won't work as expected.

This should be sufficient though.

* * *

This is desirable as an *alternative known good way* to invoke the Nix build on non-NixOS. Installing Nix globally using the upstream installer should work just as well. Using the distro-provided packages is hit-and-miss, due to unclear reasons, depending on the distro.

*In addition*, it may be useful for macOS or Windows users using *whatever cursed thing docker provides for them*.

* * *

Tested on Ubuntu

image

Fixes #5

### DIFF
--- a/support/docker/build.sh
+++ b/support/docker/build.sh
@@ -55,10 +55,12 @@ if [ -e /Tow-Boot ]; then
 	nix-build --cores 0 "$@"
 	)
 
+	stderr.printf "(Unwrapping store paths...)\n"
+
 	# Unwrap the `result` symlink
 	out=$(readlink -f result)
 	rm result
-	cp -r "$out" result
+	cp -vr "$out" result >&2
 	chown -R "$REAL_UID:$REAL_GID" result
 else
 	if [[ "$relative_dir" =~ ^/ ]]; then
--- a/support/docker/build.sh
+++ b/support/docker/build.sh
@@ -45,23 +45,29 @@ if [ -e /Tow-Boot ]; then
 	REAL_UID="$1"; shift
 	REAL_GID="$1"; shift
 
-	# Cleanup a leftover result (unclean and rude)
-	if [ -e result ]; then
-		rm -rf result
-	fi
+	# It's fine to use a static name here since it's an ephemeral container.
+	mkdir -p "/tmp-build-dir"
+	cd "/tmp-build-dir"
 
 	# Call `nix-build`, showing the user the actual invocation.
-	(set -x
-	nix-build --cores 0 "$@"
+	(
+	NIX_PATH="Tow-Boot=/Tow-Boot"
+	set -x
+	nix-build \
+		--cores 0 \
+		'<Tow-Boot/default.nix>' \
+		"$@"
 	)
 
-	stderr.printf "(Unwrapping store paths...)\n"
+	stderr.printf "\n(Unwrapping store paths...)\n"
 
-	# Unwrap the `result` symlink
-	out=$(readlink -f result)
-	rm result
-	cp -vr "$out" result >&2
-	chown -R "$REAL_UID:$REAL_GID" result
+	for f in *; do
+		target="$(readlink -f "$f")"
+		rm -rf "$current_dir/$f"
+		stderr.printf "%s -> %s\n" "$f" "$target"
+		cp -r "$target" "$current_dir/$f"
+		chown -R "$REAL_UID:$REAL_GID" "$current_dir/$f"
+	done
 else
 	if [[ "$relative_dir" =~ ^/ ]]; then
 		stderr.printf "Error: the docker wrapper script needs to be executed in the Tow-Boot checkout.\n"
--- a/support/docker/build.sh
+++ b/support/docker/build.sh
@@ -21,13 +21,22 @@ set -u
 set -o pipefail
 PS4=" $ "
 
+project_dir="$(cd "$(dirname "$0")/../.."; pwd)"
+current_dir="$(pwd)"
+relative_dir="${current_dir/$project_dir/.}"
+
+stderr.printf() {
+	printf "$@" >&2
+}
+
 _nix_docker() {
 	local STORE_VOLUME
 	STORE_VOLUME="$(docker volume create "Tow-Boot--nix-store")"
 
 	docker run \
 		--volume "$STORE_VOLUME:/nix" \
-		--volume "$(cd "$(dirname "$0")/../.."; pwd):/Tow-Boot" \
+		--volume "$project_dir:/Tow-Boot" \
+		--workdir "/Tow-Boot/$relative_dir" \
 		-it nixos/nix \
 		"$@"
 }
@@ -35,8 +44,6 @@ _nix_docker() {
 if [ -e /Tow-Boot ]; then
 	REAL_UID="$1"; shift
 	REAL_GID="$1"; shift
-
-	cd /Tow-Boot
 
 	# Cleanup a leftover result (unclean and rude)
 	if [ -e result ]; then
@@ -54,6 +61,11 @@ if [ -e /Tow-Boot ]; then
 	cp -r "$out" result
 	chown -R "$REAL_UID:$REAL_GID" result
 else
+	if [[ "$relative_dir" =~ ^/ ]]; then
+		stderr.printf "Error: the docker wrapper script needs to be executed in the Tow-Boot checkout.\n"
+		exit 1
+	fi
+
 	# Outside the docker environment, we're calling this script in the container.
 	_nix_docker /Tow-Boot/support/docker/build.sh \
 		"$(id -u)" "$(id -g)" \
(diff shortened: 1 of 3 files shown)


support/docker/build.sh
```diff
@@ -0,0 +1,61 @@
...
+set -e
+set -u
+set -o pipefail
+PS4=" $ "
+
+_nix_docker() {
+	local STORE_VOLUME
+	STORE_VOLUME="$(docker volume create "Tow-Boot--nix-store")"
+
+	exec docker run \
```
lheckemann: Using `exec` in a function can be confusing, since that causes it not to return. That doesn't make a difference in this scenario, but if the script ever does more outside docker it could get confusing. Maybe remove that here?

support/docker/build.sh
```diff
@@ -0,0 +1,61 @@
...
+		--volume "$STORE_VOLUME:/nix" \
+		--volume "$PWD:/Tow-Boot" \
```
lheckemann: This assumes that the user is currently in the Tow-Boot root directory.

```suggestion
		--volume "$(cd "$(dirname "$0")/../.."; pwd):/Tow-Boot" \
```
While this also makes an assumption (that the script hasn't been moved elsewhere), I think that's an assumption that's more likely to hold.

support/docker/build.sh
```diff
@@ -0,0 +1,61 @@
...
+
+	# Unwrap the `result` symlink
+	out=$(readlink -f result)
+	rm result
+	cp -r "$out" result
+	chown -R "$REAL_UID:$REAL_GID" result
+else
+	# Outside the docker environment, we're calling this script in the container.
+	_nix_docker /Tow-Boot/support/docker/build.sh \
+		"$(id -u "$USER")" "$(id -u "$USER")" \
```
lheckemann:
```suggestion
		"$(id -u "$USER")" "$(id -g "$USER")" \
```
or potentially only
```suggestion
		"$(id -u)" "$(id -g)" \
```
It's quite unclear to me where the environment variable USER comes from (it's certainly not bash) and I think using the IDs that we're running as is more likely to behave as desired than a user database lookup.

support/docker/build.sh
```diff
@@ -0,0 +1,61 @@
...
+		rm -rf result
+	fi
+
+	# Call `nix-build`, showing the user the actual invocation.
+	(set -x
+	nix-build --cores 0 "$@"
+	)
+
+	# Unwrap the `result` symlink
+	out=$(readlink -f result)
```
lheckemann: Capturing stdout of `nix-build` could be better than doing this, not reliant on only one derivation being built and `--no-out-link` or `-o` not being passed. I think it's also less racy?

support/docker/build.sh
```diff
@@ -0,0 +1,61 @@
...
+	rm result
+	cp -r "$out" result
+	chown -R "$REAL_UID:$REAL_GID" result
+else
+	# Outside the docker environment, we're calling this script in the container.
+	_nix_docker /Tow-Boot/support/docker/build.sh \
```
lheckemann, line 78: Personally I like putting "environment setup" at the top of the script, checking if we're in the environment we want (`[ -e /Tow-Boot ]` in this script) and `exec`ing if not, then having the material part of the script outside any conditions. That's just a matter of taste though :)

lheckemann (approved):

support/docker/build.sh
samueldr: I'm concerned about multiple outputs though. And yes I hate that part.

If you have any suggestions, I'm all ears.

support/docker/build.sh
samueldr: Yeah, I even had the "debate" internally... I wanted to get rid of the needless parent shell process... But at what cost?

support/docker/build.sh
samueldr: As you said ¯\\\_(ツ)\_/¯.

In more complex situations I would implement more bash functions and at the end of the script would be the tiniest amount of code to "route" to the right functions. But this is of a relatively manageable size considering the magic it does behind the scenes.

samueldr:
Other than the current (1) unresolved discussion, I believe this should be ready.

At worst, the unresolved discussion would be something to consider in re-evaluating the script later down the line.

support/docker/build.sh
lheckemann: How about building outside the `/Tow-Boot` volume (so the directory where the result symlink ends up is owned purely by the container) and then copying everything that ends up in said directory to `/Tow-Boot` afterwards? This would further break passing paths (currently it behaves as one would expect if the script is invoked from the tow-boot checkout) but I'm questioning whether we should really pass the arguments to `nix-build` directly at all --- I feel like too much of the interface is broken by this script and allowing the user to pass only attrs from the top level of Tow-Boot might be better than passing everything through as is to `nix-build`.

support/docker/build.sh
```diff
@@ -0,0 +1,61 @@
...
+	# Call `nix-build`, showing the user the actual invocation.
+	(set -x
+	nix-build --cores 0 "$@"
+	)
+
+	# Unwrap the `result` symlink
+	out=$(readlink -f result)
+	rm result
+	cp -r "$out" result
+	chown -R "$REAL_UID:$REAL_GID" result
```
lheckemann: Will the copy always be fast enough that reporting what's happening isn't necessary?

support/docker/build.sh
samueldr: I wonder too. A goal I tried to achieve is that it should map as closely to Nix as possible, such that upgrading to using Nix proper just needs to change the invocation to `nix-build`.

support/docker/build.sh
samueldr: No idea. Outputs could be big, computers slow.

support/docker/build.sh
lheckemann: In that case, the script should probably respect the user's cwd. Passing a path will, in the current incarnation, only work as expected if the user _happens_ to be in the root of the Tow-Boot repo.

support/docker/build.sh
lheckemann: Then it should probably report that it's copying the output --- since Nix itself has finished at this point, it looks like the whole process could be complete and the outputs "ready to use".

support/docker/build.sh
samueldr: *except* we're in the docker container, which means we can't know the actual CWD.

This could be done, though, by detecting the CWD relative to the root of the project, and only accept incantations inside the repo.

lheckemann: Exactly, we're already wrapping docker in this shell script, so we can also work out where we are relative to the repo root and pass it using `docker run`'s `-w`.

samueldr: I think we have to sacrifice at least one part of the CLI for this to work without parsing the whole CLI.

We *can* build in another directory to easily expand all result links, but `./.` now refers to somewhere else, and thus `error: getting status of ...: No such file or directory`.

Given I prefer expanding all out links to allow building arbitrary `.nix` files, I think the concession will be to forcibly refer to `<Tow-Boot/default.nix>`. Anyone needing something more involved is free to edit the script, install Nix or use some other means. At least *everything else* on the CLI should be fine, other than referring to additional paths. Which is fine, since I dropped supporting paths for `--arg device` in #89.

> I'm using the angled brackets here to hide the implementation detail of the volume... Otherwise `/Tow-Boot/default.nix` would be used, which at this point basically has as much value.

If you have any ideas how to have our cake and eat it too (allow arbitrary relative paths to still work), I'm all ears.

support/docker/build.sh
samueldr: It's now quite obvious.

image

samueldr:
- Updated with current state of `development`
 - Made the shim honor project-relative cwd
 - Made the script unwrap all output paths
 - Updated documentation

support/docker/build.sh
lheckemann: Fair enough. This was mostly about making sure the limitations are a conscious decision and not an unfortunate accident :)

>have our cake and eat it too

We could mount the entire host system into the container. But that would be quite rude.

samueldr: It wouldn't solve identifying new build outputs to unwrap! With the current strategy, the build needs to happen in an empty directory, such that we can assume `*` represents all out paths.

The issue is on the CLI `./.` is CWD, and cannot be changed, so there is no way to have an empty directory, and still having `./.` refer to somewhere else.

Which is why it now forcibly builds `<Tow-Boot/default.nix>`

PR #85, opened 2022-01-30
radxa-RockPi4: Add support for RockPi 4 model A/B

CRTified:
# Add support for RockPi 4 model A/B

Also deduplicate the configuration of RockPi 4C, which is closely
related to the 4A/B.
For a short overview:

 - Model A: No WiFi, no DP-out
 - Model B: WiFi, no DP-out
 - Model C: WiFi, DP-out

The model A used to test this config reports itself as model B.
As the difference in the device trees of A/B only consists of
additional SPIO/UART nodes for WiFi/BT, it likely will work on both
models without any problems.

The GPIO pins used for WiFi/BT are left unconnected and are not routed
anywhere else according to the schematics available at [1].

[1] https://wiki.radxa.com/Rockpi4/hardware

# Patch for `development`

If anyone wants to build the Image on the current state of the `development`-branch, here's an additional patch (apply with `patch -p1`):
```patch
diff --git a/boards/radxa/default.nix b/boards/radxa/default.nix
index cff959d..6f471fb 100644
--- a/boards/radxa/default.nix
+++ b/boards/radxa/default.nix
@@ -1,6 +1,19 @@
 { rockchipRK399 }:
 
 {
+  radxa-RockPi4 = rockchipRK399 {
+    boardIdentifier = "radxa-rockPi4";
+    defconfig = "rock-pi-4-rk3399_defconfig";
+    SPISize = 4 * 1024 * 1024; # 4 MiB
+    patches = [
+      # Based on https://github.com/armbian/build/blob/master/patch/u-boot/u-boot-rockchip64/board-rock-pi-4-enable-spi-flash.patch
+      ./0001-rockpi4-rk3399-add-spi-support.patch
+      # From https://github.com/armbian/build/blob/master/patch/u-boot/u-boot-rockchip64/general-add-xtx-spi-nor-chips.patch
+      ./general-add-xtx-spi-nor-chips.patch
+    ];
+    withSPI = true;
+  };
+
   radxa-RockPi4C = rockchipRK399 {
     boardIdentifier = "radxa-RockPi4C";
     defconfig = "rock-pi-4c-rk3399_defconfig";
```


samueldr:
(The base branch shenanigans were done to force GitHub to refresh the change set... might not have been needed if I was more patient.)

Looks good!

PR #89, opened 2022-02-06
Go back to the previous build UX

samueldr:
I agree that `--argstr device ...` is inconvenient. Let's roll back the same UX as was used previously, but this time using the new eval.

cc @colemickens


samueldr:
Before I forget, paraphrased from the matrix room

> `pkgs` should be an input again.

And yes, we can do that.

samueldr:
@colemickens: added back `pkgs`, please tell if that's as you expected.

PR #91, opened 2022-02-07
patches: Revert RK3399 boot order re-ordering from upstream

### LINKED ISSUE #86: regression: In release 004, SD has priority over eMMC with RK3399
Unverified if the order is wrong for other SOCs too.

samueldr:
Fixes #86.

### DIFF
--- a/support/u-boot/2021.10/patches/0001-Revert-rockchip-Fix-MMC-boot-order.patch
+++ b/support/u-boot/2021.10/patches/0001-Revert-rockchip-Fix-MMC-boot-order.patch
@@ -1,0 +1,35 @@
+From ba65875dc7311d251f2e4c12179c9a589e997477 Mon Sep 17 00:00:00 2001
+From: Samuel Dionne-Riel <samuel@dionne-riel.com>
+Date: Sun, 6 Feb 2022 20:16:55 -0500
+Subject: [PATCH] Revert "rockchip: Fix MMC boot order"
+
+This reverts commit b212ad24a604b00b240add35516b7381965deb31.
+
+This "fix" goes against the preferred boot order for Tow-Boot, which
+prefers internal storage first.
+---
+ include/configs/rockchip-common.h | 6 +++---
+ 1 file changed, 3 insertions(+), 3 deletions(-)
+
+diff --git a/include/configs/rockchip-common.h b/include/configs/rockchip-common.h
+index ba7061a287c..0b9e24d1db4 100644
+--- a/include/configs/rockchip-common.h
++++ b/include/configs/rockchip-common.h
+@@ -14,11 +14,11 @@
+ 
+ #ifndef CONFIG_SPL_BUILD
+ 
+-/* First try to boot from SD (index 1), then eMMC (index 0) */
++/* First try to boot from SD (index 0), then eMMC (index 1) */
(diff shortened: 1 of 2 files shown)


support/u-boot/2021.10/patches/0001-Revert-rockchip-Fix-MMC-boot-order.patch
```diff
@@ -0,0 +1,35 @@
+From ba65875dc7311d251f2e4c12179c9a589e997477 Mon Sep 17 00:00:00 2001
+From: Samuel Dionne-Riel <samuel@dionne-riel.com>
+Date: Sun, 6 Feb 2022 20:16:55 -0500
+Subject: [PATCH] Revert "rockchip: Fix MMC boot order"
+
+This reverts commit b212ad24a604b00b240add35516b7381965deb31.
+
+This "fix" goes against the preferred boot order for Tow-Boot, which
+prefers internal storage first.
```
a-wai, line 9: This raises a question: how can one recover from a broken install on the internal storage? For example, if the kernel can boot but panics before starting `init`, this would make booting from a recovery SD impossible without using UART.

samueldr: There are two orders to consider:

 * Platform Firmware Startup order
 * Operating System Boot order

This is about the latter, the OS boot order.

As long as the platform firmware works, the OS boot order is always trivial-enough to change via user input.

Note that Tow-Boot is usable on many platforms, it is not specific to a particular type of platform, or a specific device.

With all supported platforms, user input is available to change the OS Boot order, even without serial. Though for some (SBCs mainly) it will require a keyboard/display if serial is not available to do so nicely.

The **default** OS boot order in Tow-Boot should always prefer an internal boot source. To boot an external boot source (SD, USB, other), user input must be made. Some would call it a security measure; it is not. It is purely a user experience issue.

Note that I said *default* OS boot order. This is because *down the line*, once things progress to the desired goals, it is intended to provide an interface to select your preferred OS boot order, like you would with a boring old BIOS.

**If** things are dire enough that boot is impossible, it is assumed that local user input will be necessary, so this wouldn't make things any more awkward than they would be either way.

* * *

As for the *platform firmware startup order*, this is a mostly immutable property of the paltform, defined in what is commonly called the "maskrom" code for the SoC. The upstream change, and this revert, change nothing about this. Before/after either change/revert, the situation is the same.

a-wai: Got it, thanks for the detailed answer!

> With all supported platforms, user input is available to change the OS Boot order, even without serial

This is the point I wasn't sure about after a quick look at the existing patches, but after looking into more details (including the phone/tablet PR) it makes sense.

samueldr: *If* a device without user input facilities was contributed to Tow-Boot, it would be warranted to re-examine the position for those boards... But it might be that it would not be fit for being *merged* in Tow-Boot proper, or if it is, kept excluded from releases until it is made possible to use *like a boring old computer/phone*.

PR #96, opened 2022-02-09
doc: Use *platform firmware*

### LINKED ISSUE #58: Replace "initial boot firmware" wording with "platform firmware"
The latter (platform firmware) is what's being used more often than not in the EBBR documentation. Let's help this wording take its righteous place.

Note that dropping "boot" from here will probably help reducing confusion about bootloader vs. platform firmware!

(Side-note for me: fix this also at NixOS on ARM on the NixOS wiki...)

samueldr:
Fixes #58.

The EBBR spec uses the term *platform firmware* enough to warrant using this instead of trying to coin a confusing new term.

### DIFF
--- a/doc/FAQ.md
+++ b/doc/FAQ.md
@@ -24,8 +24,8 @@ First, the "any `AArch64` devices" has to assume that the operating system can
 run on those devices. For this question, we assume the operating system does.
 
 It all depends on whether the systems require dedicated or shared storage for
-the firmware. If the system requires shared storage, and can only boot the
-initial firmware from the SD card, the SD card must contain its firmware. This
+the firmware. If the system requires shared storage, and can only start the
+platform firmware from the SD card, the SD card must contain its firmware. This
 will probably not work well for two different enough platforms.
 
 Though, still assuming the operating system can boot on all the devices you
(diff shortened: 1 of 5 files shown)


doc/FAQ.md
```diff
@@ -24,8 +24,8 @@ First, the "any `AArch64` devices" has to assume that the operating system can
 run on those devices. For this question, we assume the operating system does.
 
 It all depends on whether the systems require dedicated or shared storage for
-the firmware. If the system requires shared storage, and can only boot the
-initial firmware from the SD card, the SD card must contain its firmware. This
+the firmware. If the system requires shared storage, and can only start the
```
samueldr, line 27: snuck a `s/boot/start` in addition, since it was obvious when fixing the wording.

Why? You *boot an OS*, you *start a platform firmware*.

Why? Because otherwise you overload the word *boot* and it can get confusing when describing things to end-users.

doc/glossary.md
```diff
@@ -27,7 +27,7 @@ Boot Loader
 -----------
 
 The *Boot loader* is a program which may or may not be distinct from the
-*Initial Boot Firmware*, which is meant to load a start an *Operating System*.
+*Platform Firmware*, which is meant to load a start an *Operating System*.
```
ItsHarper:
```suggestion
*Platform Firmware*, which is meant to load and start an *Operating System*.
```

samueldr: Good catch, thank you!

PR #97, opened 2022-02-11
tow-boot/builder: Fix Rockchip DTS erroneous 15200 baud rate

samueldr:
This did not end up causing issues because the other large-scale
replacement was the one that took precedence. This one *may* have caused
issues with `/chosen/stdout` though.

Thank you @Danct12 for noticing.

### DIFF
--- a/modules/tow-boot/builder.nix
+++ b/modules/tow-boot/builder.nix
@@ -131,7 +131,7 @@ in
             done
             for f in arch/arm/dts/*rk3399*.dts* arch/arm/dts/*rk3328*.dts*; do
               (set -x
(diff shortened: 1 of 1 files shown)


Danct12:
> Thank you @Danct12 for noticing.

My pleasure! I'd also thank you for making an awesome bootloader.

PureTryOut:
> an awesome bootloader.

@Danct12 I know you use u-boot as a bootloader too, but let's start calling tow-boot and u-boot "platform firmware" to avoid confusion. Tow-Boot does _not_ replace bootloaders like GRUB or systemd-boot of course :wink:

Danct12:
OK

Issue #92, opened 2022-02-07
Revise all OS boot orders


samueldr:
Turns out I misremembered what I implemented when I stated the more general rule "internal first, external last"...

 - https://github.com/Tow-Boot/Tow-Boot/blob/eb2a83466e220e31b561955b2c59b9fc9c21cd46/support/builders/tow-boot/patches/0001-Tow-Boot-sunxi-ignore-mmc_auto-force-SD-then-eMMC.patch#L4

So this is considered a bug, especially considering #86.

* * *

Families:

 -  ⚫ Raspberry Pi (Not applicable)
 -  ✔️ Allwinner (#101)
 -  ✔️ Rockchip (#91)
 -  ✔️ Amlogic (#98)

PR #98, opened 2022-02-11
radxa-zero2: Add support for Radxa Zero 2

samueldr:
This adds support for the [Radxa Zero 2](https://wiki.radxa.com/Zero2).

This also adds the required Amlogic A311D support.

This family of board has *three* possible `MMC[0-2]` values it could be using, thus the support for naming MMC interfaces had to be updated to allow naming `MMC2`.

Thus this change also shows how to edit the naming of MMC interfaces for boards having weirder setups.

* * *

At this point in time, *mmc boot* support is not implemented, but it is coming in a further WIP change. MMC boot support is not strictly required for adding the port to Tow-Boot.

This can currently be used from eMMC or SD card.

### Tested:

 - HDMI output
 - USB input (seems to exhibit the same issues as other amlogic boards, which is good I guess. Namely, some "menu input events" may need to be "pumped out" by pressing CTRL a couple of times if it seems to not respond.)
 - Synced amlogic DTs with mainline, though other than "improved" 5.16 compatibility, I expect nothing really matters.
 - The latest NixOS UEFI "new kernel" iso boots, though it seems there might be some form of graphical issues... Highly likely to not be relevant to Tow-Boot.
     - (Slightly unrelated: Installed NixOS UEFI iso with "current LTS" (5.10) stock and it is alright, no weird rendering issues. Though it shows the innate FDT is alright.)

* * *

5117f6b73cd988ef14d20e575eb41a09a98c2a93 Handles #92 for Amlogic (hopefully correctly for all vintages... otherwise we'll need to make this configurable.)

* * *

cc @RadxaYuntian

### DIFF
--- a/boards/radxa-zero2/0001-radxa-zero2-board-enablement.patch
+++ b/boards/radxa-zero2/0001-radxa-zero2-board-enablement.patch
@@ -1,0 +1,690 @@
+From d98ead879ae1d6bc168e59dab3ec53a649d5518e Mon Sep 17 00:00:00 2001
+From: Yuntian Zhang <yt@radxa.com>
+Date: Fri, 14 Jan 2022 19:16:39 +0800
+Subject: [PATCH 1/3] boards: amlogic: add Radxa Zero 2 defconfig
+
+Add a defconfig for the Radxa Zero 2 SBC, using Amlogic A311D SoC.
+This file is created manuall based on radxa-zero_defconfig and
+khadas-vim3_defconfig.
+
+Signed-off-by: Yuntian Zhang <yt@radxa.com>
+(cherry picked from commit 1df0b2a2f182348ee440aea62c4d9b5776d4e1c7)
+---
+ configs/radxa-zero2_defconfig | 93 +++++++++++++++++++++++++++++++++++
+ 1 file changed, 93 insertions(+)
+ create mode 100644 configs/radxa-zero2_defconfig
+
+diff --git a/configs/radxa-zero2_defconfig b/configs/radxa-zero2_defconfig
+new file mode 100644
+index 00000000000..b4cbc5c37dc
+--- /dev/null
++++ b/configs/radxa-zero2_defconfig
+@@ -0,0 +1,93 @@
++CONFIG_ARM=y
++# CONFIG_SYS_BOARD is not set
++CONFIG_ARCH_MESON=y
++CONFIG_SYS_TEXT_BASE=0x01000000
++CONFIG_NR_DRAM_BANKS=1
++CONFIG_ENV_SIZE=0x2000
++CONFIG_DM_GPIO=y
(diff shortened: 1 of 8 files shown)


RadxaYuntian:
@spikerguy stuff is happening.

boards/radxa-zero2/0001-radxa-zero2-board-enablement.patch
```diff
@@ -0,0 +1,690 @@
...
++		serial0 = &uart_AO;
++	};
++
++	chosen {
++		stdout-path = "serial0:115200n8";
++	};
++
++	memory@0 {
++		device_type = "memory";
++		reg = <0x0 0x0 0x0 0x80000000>;
```
samueldr, line 199: `0x80000000` bytes is 2GiB, and indeed in the OS I see 2GiB available.

It seems there is no SKUs with less than 4GiB listed in the SKUs list. Shouldn't this be `0x100000000`?

*Z32M* on what I presume is the RAM chip makes me think that yes it's 4GiB.

Alternatively, I guess this could be detected during firmware startup, but it seems `meson` in U-Boot doesn't have facilities for this, or I'm not seeing it. AFAICT it only will use the `/memory` node from its built-in FDT blob. (`dram_init` in `arch/arm/mach-meson/board-common.c`).

EDIT: could be `get_effective_memsize` in `arch/arm/mach-meson/board-g12a.c` that would normally handle that. I will have some reading to do about amlogic.

RadxaYuntian: That should be a 2GB chip. We actually have 1 board fitted with 4GB memory for Android, and when I tested the same dtb, U-Boot and Linux were able to detect the full size of memory. Rest of the boards are using 2GB memory since this is just some samples.

Also integers used there are 32 bits, so if you want to specify 4GB you need to do `reg = <0x0 0x0 0x1 0x0>;`.

samueldr: So Z32M-G 2052 here is not 32 megabits (`32 / 8 → 4`)? Wrong assumption on my end. I couldn't find specs, or even a sales sheet listing that product code. (Though I suspect 2052 could be a date code.)

With the testing you've done, assuming indeed the board I have is fitted with 2GiB, then there's no issue I figure.

I still wonder: should that node represent the amount of RAM the end-user product is expected to use?

samueldr: False alarm. Got it, tested things by reducing the hardcoded value to 1GiB, and yeah, everything is fine.

I guess what confused me was the combination of:

 - Not being aware that it had 2GiB instead of 4GiB of RAM
 - The *Z32M* marking which I assumed meant 32 gigabits
 - Coincidentally the value in the DT was 2GiB

PR #99, opened 2022-02-12
Add Odroid N2+ support

Strit:
The Odroid N2+ is almost the same device as the regular N2, except that it
supports some other clock speeds.

So change the default dts to the N2+ one so we can make use of those.

While N2 dts works on the N2+, an N2+ dts can be dangerous on the N2.

PS: I don't know how Nix scipts work, but I tried just copying from the Odroid N2 board folder and make the slight change that was needed there. If something is wrong please let me know.

### DIFF
--- a/boards/odroid-N2-plus/0001-Enable-the-SPI-on-the-ODROID-N2-by-default.patch
+++ b/boards/odroid-N2-plus/0001-Enable-the-SPI-on-the-ODROID-N2-by-default.patch
@@ -1,0 +1,57 @@
+From 367dfbebd1099f7c3d58ee1da007653f34fb0f69 Mon Sep 17 00:00:00 2001
+From: Las <las@protonmail.ch>
+Date: Wed, 16 Jun 2021 12:28:17 +0000
+Subject: [PATCH] Enable the SPI on the ODROID N2 by default
+
+This halves the the bandwidth to the eMMC, but given
+that this will apparently be restored when booting a kernel with the
+default device tree, it isn't really important.
+---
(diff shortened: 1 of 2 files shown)


samueldr:
Was the build tested? I haven't looked yet, but reading "but I tried just copying from" this could go either way.

I merged recently [a docker-using build shim](https://github.com/Tow-Boot/Tow-Boot/pull/83), which can be used for people that don't want to or can't install Nix globally.

* * *

How is "N2 vs. N2+" intended to be solved at upstream U-Boot? Any discussions you are aware of?

Though looking quickly, it seems the defconfig is meant for either:

```
CONFIG_IDENT_STRING=" odroid-n2/n2-plus"
```

Furthermore, it looks like since 2021.10, [detection should now provide the right fdtfile suffix](https://source.denx.de/u-boot/u-boot/-/commit/480aeb0aa7b1fcb78e15d1678610f61ff421b883).

So uh, it looks like it might already be working out of the box using the N2 build, for 2021.10+ builds.

Strit:
I haven't build tested it, no.

Regarding upstream uboot. I didn't know that they had any detection built in for it. If that's the case and the "plus" dtb does get used, then this can be closed as it's not needed.

Strit:
As my Odroid N2+ cannot seem to communicate with the SPI flash, even though it detects it, I can't verify if the regular Odroid N2 version uses loads the N2+ dts or not.

So lets close this for now, assuming that the board detection in u-boot works as it should. We can always re-open this if we find it is not the case.

PR #100, opened 2022-02-12
Doc: Distributions: Add Manjaro ARM

Strit:
This is the first iteration of documentation on how to install Manjaro on a device with Tow-Boot.

So far only Tow-Boot on SPI is described and UEFI not supported yet.

### DIFF
--- a/doc/distributions/manjaro.md
+++ b/doc/distributions/manjaro.md
@@ -1,5 +1,5 @@
 Manjaro
------
+=======
 
 Notes specific to Manjaro.
 
--- a/doc/distributions/manjaro.md
+++ b/doc/distributions/manjaro.md
@@ -1,0 +1,20 @@
+Manjaro
+-----
+
+Notes specific to Manjaro.
+
+## Installation of Manjaro on a device with Tow-Boot on SPI
+
+This guide only covers this use case for now, as the regular Manjaro ARM images for the devices they support, already ships with U-boot on the pre-made image.
+
+Booting Manjaro on a device that already has Tow-Boot on the SPI flash is really easy.
+
+While any of the device specific images should still boot just fine with Tow-Boot, Manjaro recently started created **Generic Aarch64** images for testing with SPI installations.
+
+These can be found on [github](https://github.com/manjaro-arm/generic-images/releases) at the moment, since it's still in early development and not many devices have been tested yet.
+
+The generic image uses Extlinux to boot the kernel and, as such, is not UEFI capable.
+
+## Installation of Manjaro via UEFI
+
+Currently not supported.
--- a/doc/distributions/manjaro.md
+++ b/doc/distributions/manjaro.md
@@ -3,18 +3,10 @@ Manjaro
 
 Notes specific to Manjaro.
 
-## Installation of Manjaro on a device with Tow-Boot on SPI
+## Supported booting mechanisms
 
-This guide only covers this use case for now, as the regular Manjaro ARM images for the devices they support, already ships with U-boot on the pre-made image.
+The Manjaro ARM Generic images only supports booting with the Extlinux schema at this stage.
 
-Booting Manjaro on a device that already has Tow-Boot on the SPI flash is really easy.
+See [instructions on Manjaro wiki](https://wiki.manjaro.org/index.php/Manjaro-ARM#Preparing_the_SPI_.28optional.29)
 
-While any of the device specific images should still boot just fine with Tow-Boot, Manjaro recently started created **Generic Aarch64** images for testing with SPI installations.
-
-These can be found on [github](https://github.com/manjaro-arm/generic-images/releases) at the moment, since it's still in early development and not many devices have been tested yet.
-
-The generic image uses Extlinux to boot the kernel and, as such, is not UEFI capable.
-
-## Installation of Manjaro via UEFI
-
-Currently not supported.
+Booting via UEFI is not supported yet, but is being worked on.
--- a/doc/installing-an-operating-system.md
+++ b/doc/installing-an-operating-system.md
@@ -17,5 +17,6 @@ Linux
 <!-- This list is **alphabetically ordered** -->
 
  - [Fedora](distributions/fedora.md)
+ - [Manjaro](distributions/manjaro.md)
  - [NixOS](distributions/nixos.md)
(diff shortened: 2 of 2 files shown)


samueldr:
This is good enough; any "standard enough" OS boot supported by U-Boot upstream is supported by Tow-Boot.

I'll give a try on some "fun" hardware I have lying around to see how well this fares.

PureTryOut:
Wait, do we really want documentation for every single distribution out there in Tow-Boot docs? That seems like the wrong place, why not keep that to the distros own docs?

samueldr:
@PureTryOut we want general notes about the state of things. There shouldn't be general installation steps or guides, and rather it should link to upstream docs.

We want this because having a distro listed under here means "look, it was tested at some point and worked".

This is also needed due to some projects not necessarily working directly with us. So if there are quirks or problems the user will face, this will allow them to have a solution. Notes about quirks are much more likely on devices using the shared storage strategy,

For example, for Fedora I noted two pain points and their workarounds (one which is now resolved). And I'm noting that even if I had a way to interact with Fedora peeps, it's likely they wouldn't have been receptive in adding details about the quirks to their docs.

As an unseen example, I probably should have added Ubuntu and Debian pages stating that they can't complete the installation process (for similar reasons, but two entirely different reasons). No known workarounds for them.

If you look at the NixOS notes, they refer to the unofficial wiki, which officially holds the notes for ARM stuff, where there is an in-depth write-up. This is the kind of docs I would hope to see for most distros.

There is value in having an almost empty page stating "eh, it works!". It signals to end-users that yes, Tow-Boot **and** that distro were tested.

* * *

So uh, good catch, maybe this should be reduced some, and maintained at an upstream location, @Strit. The Tow-Boot doc page probably should be reviewed, maybe stating something along the lines that support is in progress? Or just stating the location of upstream docs, and support is available through the `extlinux.conf`-compatible boot scheme. UEFI not being supported at this time.

I'm also thinking that these pages should always consider the *totally unfun* shared storage strategies using the upstream installers. Even if only stating "using the normal installation steps works fine". I could see the NixOS docs mentioning something like:

> When installing to the same media with a shared disk image, remember to keep the existing partition table, and existing partitions. Don't start from a blanked partition table.

And @PureTryOut, similarly I'd like to see a postmarketOS doc, referring to an upstream documentation source owned by postmarketOS, and describing how it's supported.

Strit:
Yeah, you are probably right. We don't have any docs for it on our end yet though, but I can link them in when we do get them.

I'll work on re-wording this PR a bit.

PureTryOut:
In case of distro specific quirks I can see _some_ value in it (although even then I think upstream should have it's own docs for it but meh), this PR is just "everything works, in fact, why are you even looking at these useless docs?" (not to talk anyone down of the effort put into it :see_no_evil: ).

Similarly I have no intention of making pmOS docs as the point is that it "just works". In cases of devices using shared storage mode we'll make sure to include them in our own device-specific installation instructions, no need to duplicate it here. If booting is really _that_ boring, there is no need for distro-specific docs :wink:

samueldr:
I think we're mostly in agreement. Quirks are for distros that are not receptive to direct collaboration, so indeed you're right it should go upstream, and even be fixed.

But, I think there is one thing you've not considered: **signalling** that a distribution supports standards-based boot (or is otherwise compatible). By listing the "trivial" distros in Tow-Boot, it creates pressure on other distributions by showing "look, all those other distros can do it, so why not you?".

Additionally, it creates a shortcut to the upstream documentation. For some distros (probably doesn't apply to postmarketOS) finding the right documentation sometimes is a real chore :).

PureTryOut:
Maybe better make a list somewhere with "Tow-Boot has been tested with the following distros: - postmarketOS, - Manjaro", etc? An entire new page just to say "yes we tested it, it works" seems a bit much.

> For some distros (probably doesn't apply to postmarketOS) finding the right documentation sometimes is a real chore :).

That seems like something for the distribution in question to fix and not something we should deal with here.

samueldr:
Lists can get unwieldy when you need to provide details. Pages are cheap enough to make.

We cannot fix all distros. Linking to upstream docs is also cheap. This way the Tow-Boot documentation becomes a sort of repository of all the upstream documentations and relevant links about EBBR, standards-based boots, or even "less standard" distro boot commands. Having a centralized location to point at will be worth it when distros and maintainers need to be approached.

It **would** be better to have an actually independent from firmware and independent from distro "taskforce" maintaining that repository of knowledge (and some other relevant knowledge). But as I've been told beforehand, (paraphrased) "it won't happen, no one wants that". So I'm starting here with this, and if/when such a taskforce ends up being desired, we can migrate and refer to it.

doc/distributions/manjaro.md
```diff
@@ -0,0 +1,12 @@
+Manjaro
+-----
```
samueldr:
```suggestion
=======
```

Page titles need to be `<h1>`, using dashes under a title makes it a `<h2>`.

doc/installing-an-operating-system.md
```diff
@@ -18,3 +18,4 @@ Linux
 
  - [Fedora](distributions/fedora.md)
  - [NixOS](distributions/nixos.md)
+ * [Manjaro](distributions/manjaro.md)
```
samueldr: Let's keep this list alphabetically ordered. M goes before N :).

Yes, I could put NixOS front and center, but I'm trying to make this extremely distro-agnostic. Alphabetical order helps for that.

samueldr (changes requested):
Minor changes.

In addition, can you rebase the unneeded changes out of the history? If not, no big deal for this PR.

I generally try working under the assumptions that PRs are an idealistic view of changes, like mailing lists changes.

Strit:
> In addition, can you rebase the unneeded changes out of the history? If not, no big deal for this PR.
> 
> I generally try working under the assumptions that PRs are an idealistic view of changes, like mailing lists changes.

I'm not a git wizard, so I'm not entirely sure what you mean here. Maybe we can just squash the commits when we are happy with the PR?

I'll commit your suggested changes and see.

samueldr:
What I was talking about was rebasing/squashing selectively the additional changes.

But it's not a problem not to do so, since you're no git wizard.

samueldr (approved):

PR #105, opened 2022-02-18
Support building stock U-Boot (without the main Tow-Boot features patch set)

samueldr:
This will not disable building device-specific or platform-specific changes. Those changes *anyway* should be limited to compatibility and bug fixes.

This can be used like so:

```nix
# tmp.nix
{
  Tow-Boot = {
    uBootVersion = "2022.01"; # Handles conditionals around patches, might need to futz around when building intermediary commits.
    useDefaultPatches = false; # Forces build without Tow-Boot features
    src = builtins.fetchGit /Users/samuel/tmp/u-boot/u-boot; # Uses a local checkout
  };
}
```

```
 $ NIXPKGS_ALLOW_UNFREE=1 time nix-build --arg configuration 'import ./tmp.nix' -A odroid-N2
```

cc @tpwrules who will probably like this right now.


tpwrules:
This seems to work nicely. I agree with the decision to keep Tow-Boot's device-specific patches and config.

PR #106, opened 2022-02-19
odroid-N2: increase default SPI speed so flash read/write works properly

### LINKED ISSUE #103: Flashing to Odroid-N2+ SPI fails
[odroid-log.txt](https://github.com/Tow-Boot/Tow-Boot/files/8098820/odroid-log.txt)
Tried to flash the 2021.10-003 release to SPI on 4GB Odroid-N2+.

* Wrote the installer to SD card with dd
* Booted the SD card
* Selected `Flash firmware to SPI` (from menu)

This failed with `SF: 8388608 bytes @ 0x0 Read: ERROR -110`.

Then tried `Completely erase SPI`, this failed with "No SPI flash selected. Please run `sf probe' ".

Attached the serial console log. In the log Tow-Boot detects the board as Odroid-N2 without plus - maybe that is the problem?

tpwrules:
It turns out the default `sf` flash command access speed of 1MHz prevents the onboard flash chip from working properly for whatever reason. This patch increases the speed to 52MHz (somewhat arbitrarily selected as half the 104MHz max speed of the chip) and now it can be read and written properly.

This fixes #103 . I can flash the SPI using the menu in the installer image, then boot off the SPI flash, and use that Tow-Boot to boot off an attached eMMC.

I will do a bit more testing and investigation and submit this upstream at some point.

### DIFF
--- a/boards/odroid/default.nix
+++ b/boards/odroid/default.nix
@@ -14,6 +14,7 @@
     extraConfig = ''
       CONFIG_USE_PREBOOT=y
       CONFIG_PREBOOT="usb start ; usb info"
+      CONFIG_SF_DEFAULT_SPEED=52000000
     '';
 
   };


samueldr:
I'll handle rebasing on `development`, which changes how configuration is handled.

Thanks for looking!

samueldr:
Merged manually in https://github.com/Tow-Boot/Tow-Boot/commit/1297c1bbb6ba882d5aa79f935fe1e55c243f639d.

Issue #104, opened 2022-02-18
doc: request to add device list and build instructions


tpwrules:
It would be useful to have a prominent list of the supported boards and their Tow-Boot names somewhere on the website.

I'd also like to see a reminder of the build instructions (i.e. `git clone ...`, `nix-build  -A ...`). Whenever I come back to this project it takes me too long to figure out what exactly I need to ask Nix for. Modifying the "warning: you're about to build all the boards" to have a link to the page above that lists the supported boards and names would be helpful too.

samueldr:
+1, anyway we need a list of "supported" devices, and a way to provide per-device information, not unlike what Mobile NixOS automatically does.

samueldr:
Relatively relevant:

```
~/tmp/u-boot/u-boot $ nix-shell ~/Projects/Tow-Boot/Tow-Boot -A pine64-pineA64LTS.config.Tow-Boot.outputs.firmware
```

This can be used to have a shell with the required environment.

samueldr:
*Incomplete* in #67; a device list was added.

In the future build instructions can be added. I want to think a bit more about it, as I don't want any focus on these. It's not *expected* that users will build. So I don't know how to expose that in the documentation.

PR #107, opened 2022-02-19
doc: Add initial supported device list

Strit:
Fixes the device list part of issue #104

### DIFF
--- a/doc/supported-devices.md
+++ b/doc/supported-devices.md
@@ -1,0 +1,40 @@
+# What devices do Tow-Boot support?
+
+## LibreComputer
+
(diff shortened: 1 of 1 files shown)


samueldr:
Hi,

Sorry it took so long to reply, but I was already cooking something up "better". For #67 I added per-device pages, which can be further improved on. Currently they list basic information, and generic, yet configurable, installation instructions.

* * *

With that said, I hadn't thought about adding links to vendor resources. I'm keeping this open to remind myself about it. I'll add the ressources you list here in a new PR later, and co-author you on it. It will require adding something in the website building infra, so I'm not even asking you to take a look.

Strit:
That's great. I didn't know you where already in the process of adding a list. You are free to take what you need from this.

PR #125, opened 2022-03-06
radxa-RockPi4(C): Add SPI documentation

CRTified:
This PR adds device-specific notes for the RockPi 4 A/B/C about the presence of SPI and what to do if it is not present.

Hope it's fine/correct to target the `released` branch for this type of documentation.

### DIFF
--- a/boards/radxa-RockPi4/README.md
+++ b/boards/radxa-RockPi4/README.md
@@ -1,0 +1,12 @@
+# Radxa ROCK Pi 4 model A/B
+
+## Device-specific notes
+
+According to the [upstream SPI documentation](https://wiki.radxa.com/Rockpi4/hardware/spi_flash),
+the SPI flash is only populated on board revisions V1.4 and later.
+
+While the schematics denote `W25Q64FV` is used, all verified units came
+with an `XT25F32B`.
+
+If no flash is present, the user either needs to use the shared storage strategy,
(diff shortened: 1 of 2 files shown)


boards/radxa-RockPi4/README.md
```diff
@@ -0,0 +1,12 @@
+# Radxa ROCK Pi 4 model A/B
+
+## Device-specific notes
+
+According to the [upstream SPI documentation](https://wiki.radxa.com/Rockpi4/hardware/spi_flash),
+the SPI flash is only populated on board revisions V1.4 and later.
+
+While the schematics denote `W25Q64FV` as used flash, all known instances came
```
samueldr:
```suggestion
While the schematics denote `W25Q64FV` is used, all verified units came
```

The wording was a bit weird, I'm fine with other rewrites too.

Main thing to consider:

 - "as used flash" doesn't work, I think.
 - "all known instances" is weird for a physical item.

samueldr:
Yes, targeting `released` is fine for docs update, since the site is built from that branch.

boards/radxa-RockPi4/README.md
CRTified: That's the result of non-native English :sweat_smile: Thanks, I've squashed the commits of PR and your suggestion together.

samueldr: I'm not a native english speaker either :).

PR #130, opened 2022-03-07
pine64-pinephoneA64: Update README.md

PsychoGame:
Mention about uBoot on SD-Card interfering with Tow-Boot installation, and write a simple example procedure to get rid of uBoot.

### DIFF
--- a/boards/pine64-pinephoneA64/README.md
+++ b/boards/pine64-pinephoneA64/README.md
@@ -4,6 +4,40 @@
 
 There is no display output for this device.
 
+
+## Startup order conflicts
+
+The Allwinner SoC used in the Pinephone (A64) will always prefer starting
+the platform firmware (U-Boot) from SD card when available.
+
+When using some pre-built distribution images, it may be desirable to neuter
+the U-Boot that is baked into the image.
+
+With the usual U-Boot setup for Allwinner, the easiest way to strip U-Boot
+from an eMMC or SD image is by running the following command, taking care
+to replace `[DEVICE]` with the device for the storage device.
+
+```
+ $ sudo dd if=/dev/zero of=/dev/[DEVICE] bs=8k seek=1 count=4
+```
+
+When running the command on the Pinephone (A64), *usually* the block device
+for the SD card is `mmcblk0`, and the block device for the eMMC is `mmcblk2`.
+You can use `lsblk` to look at the block devices available.
+
+> **NOTE**: Make sure the operating system image has a baked-in U-Boot
+> install before issuing the command, and that you are targeting the right
+> storage device.
+>
+> Some distributions target UEFI boot, and use GPT. Issuing this command
+> **will break the partition table**.
+
+Finally, on distributions shipping U-Boot, it is prudent to disable the
+package that provides the automatic upgrade facilities.
+
+ - For Manjaro and Arch Linux, the package is called `uboot-pinephone`.
+
+
 ## Additional features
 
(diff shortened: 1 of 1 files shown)


boards/pine64-pinephoneA64/README.md
```diff
@@ -2,6 +2,17 @@
...
 
+When using prebuilt distribution images it's wise to strip uBoot from the image.
+With the use of prebuilt images on SD-Card it's mandatory to strip uBoot from it, as uBoot will otherwise interfere with Tow-boot and only boot from SD-Card.
+To strip uBoot from a SD-Card installation I used the Gnome Disks utility, but any other should work as well.
+1. Select the "Unused Space (32MB)" at the beginning of the SD-card.
+2. Choose option "Format Partition" and make sure to select "Wipe". The filesystem doesn't really matter.
+3. After formatting you can choose to delete the partition again as well, but that's up to you.
+4. To make sure uBoot doesn't get reinstalled you need to remove the uboot package from the distro packages. (In Manjaro's case this package is called "uboot-pinephone")
+
+To strip uBoot from the eMMC installation I used above mentioned procedure, but then eMMC mounted as block device throught Tow-boot.
```
ItsHarper: This sentence has a confusing structure IMO

boards/pine64-pinephoneA64/README.md
```diff
@@ -2,6 +2,17 @@
 
 ## Device-specific notes
 
+When using prebuilt distribution images it's wise to strip uBoot from the image.
+With the use of prebuilt images on SD-Card it's mandatory to strip uBoot from it, as uBoot will otherwise interfere with Tow-boot and only boot from SD-Card.
+To strip uBoot from a SD-Card installation I used the Gnome Disks utility, but any other should work as well.
```
PureTryOut: Don't use terms like "I" in documentation, the reader has no clue who you are and neither should they care.

Also in general I would replace these steps with a general "wipe the first 32MB", maybe alongside a `dd` command which does so (since every Linux user has dd).

samueldr:
Hi!

Sorry it took so long until I had the time to look in this PR.

* * *

There is a problem with the instructions as they are: what if the distro image does not ship with a *U-Boot* install on it? Erasing a span of the disk image without knowing what's there could break the image in different ways. For example, if it's partitioned with GPT with a full GPT, this would break the GPT.

Otherwise, there is an unlikely risk of there being a "real" partition in the location.

What would be preferable is to erase the "unused" space between the partition table (be it MBR or GPT), and the first partition. In addition, it would likely be *correct* for any system, not specific to Allwinner. (And even for Allwinner, it's partly wrong since there are multiple offsets that could be used.) (Another side-note, it wouldn't handle platform firmware in protective partitions, but that's easier for the end-user to deal with in many ways.)

Having a script that can do this (with well-established tools) would be a correct solution. 

Another alternative could be to ship a specialized static binary, but this is not a "fun" operation to program for.

Thoughts?

samueldr:
(Moving to next milestone to expedite shipping the smbios update, there's still unanswered questions, and maybe the erasing questions should be answered by a tool that detects things out... A chunk of that has been written already elsewhere by someone else.)

samueldr:
Thanks again for the changes, I've taken to apply fixes rather than try and do some back-and-forth to get to a specific result.


With this force-push, I:

 - squashed the commits
 - re-ordered the section
 - added warnings
 - fixed formatting

I believe this is meaningfully the same as the previous version, though I always feel bad pushing major changes on top of someone else's work.

Did I miss anything with this update?

PR #133, opened 2022-03-10
pine64-pinephonePro: fix ALS interrupt pin

a-wai:
On this device, GPIO4_D3 is used as the interrupt pin for the
proximity/ambient light sensor. However, even though the datasheet
indicates it's configured as an input by default, the boot ROM sets this
pin as an output, causing issues when probing the sensor in linux.

This adds a patch for making sure this line is configured properly on
the PPP, and reverts some previous DT changes (those break
`dm_lookup_by_name()` on RK3399, as the driver expects the gpio node
names to include the bank number right before '@').

### DIFF
--- a/boards/pine64-pinephonePro/0001-pine64-pinephonepro-device-enablement.patch
+++ b/boards/pine64-pinephonePro/0001-pine64-pinephonepro-device-enablement.patch
@@ -235,51 +235,6 @@ index 3871c7fd83b..d3cdf6f42a3 100644
  		status = "disabled";
  
  		ports {
-@@ -1958,7 +2038,7 @@
- 		#size-cells = <2>;
- 		ranges;
- 
--		gpio0: gpio0@ff720000 {
-+		gpio0: gpio@ff720000 {
- 			compatible = "rockchip,gpio-bank";
- 			reg = <0x0 0xff720000 0x0 0x100>;
- 			clocks = <&pmucru PCLK_GPIO0_PMU>;
-@@ -1971,7 +2051,7 @@
- 			#interrupt-cells = <0x2>;
- 		};
- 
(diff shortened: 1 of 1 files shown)


samueldr:
I see the changes to the DT are from here:

 - https://lore.kernel.org/all/20211007144019.7461-3-jbx6244@gmail.com/
 - https://github.com/torvalds/linux/commit/ec3028e7c83ed03f9cd10c0373d955b489ca5ed6

Ideally we'll re-sync the whole DT source again with `pine64-org/linux`, which has that change now too.

The other patch looks alright to add.

* * *

Note that in this release cycle I want to address the issue you're kinda facing in this PR: managing the patch sets. If you have any experience/opinions, I would be glad to hear them.

a-wai:
> Ideally we'll re-sync the whole DT source again with `pine64-org/linux`, which has that change now too.

I'm a bit worried that this will lead to breakages in u-boot itself, as highlighted by the `dm_gpio*` issues here. I assume this will be solved anyway once we have the PPP DT upstream in both the kernel and u-boot, maybe some bits should be left unsync'ed?

Anyway, closing due to authorship fights.

a-wai:
> Note that in this release cycle I want to address the issue you're kinda facing in this PR: managing the patch sets. If you have any experience/opinions, I would be glad to hear them.

On that particular issue I don't have much to say, but I confess wondering whether I should make a separate patch or append it to the existing one. I opted for the latter as the file already contained numerous patches; if it were broken down into multiple patch files I'd definitely have created a separate patch.

Don't know if that helps...

samueldr:
Kind of. It helps in showing a *process* needs to be defined, at the very least, if no tooling fits the needs.

PR #135, opened 2022-03-18
pine64-pinephonePro: add smbios data

MartijnBraam:
Adds smbios data for the PinePhone Pro, tested on actual hardware with it running from SPI.

This should allow fwupd to function since it uses this data to generate unique device IDs

### DIFF
--- a/boards/pine64-pinephonePro/0001-rk3399-pinephone-pro-add-smbios-info.patch
+++ b/boards/pine64-pinephonePro/0001-rk3399-pinephone-pro-add-smbios-info.patch
@@ -1,0 +1,71 @@
+From b58fa10598ac446784615f56030150d205e96487 Mon Sep 17 00:00:00 2001
+From: Martijn Braam <martijn@brixit.nl>
+Date: Fri, 18 Mar 2022 15:48:46 +0100
+Subject: [PATCH 1/2] rk3399-pinephone-pro: add smbios info
+
+Provide the SMBIOS data for the PinePhone Pro. U-Boot still hardcodes
+the chassis type and related values.
+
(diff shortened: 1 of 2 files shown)


boards/pine64-pinephonePro/0001-rk3399-pinephone-pro-add-smbios-info.patch
```diff
@@ -0,0 +1,67 @@
...
+
+diff --git a/arch/arm/dts/rk3399-pinephone-pro-u-boot.dtsi b/arch/arm/dts/rk3399-pinephone-pro-u-boot.dtsi
+index 28bec0a6a7..1e7378a288 100644
+--- a/arch/arm/dts/rk3399-pinephone-pro-u-boot.dtsi
++++ b/arch/arm/dts/rk3399-pinephone-pro-u-boot.dtsi
+@@ -19,6 +19,22 @@
+ 	config {
+ 		u-boot,spl-payload-offset = <0x80000>; /* @ 512KB */
+ 	};
++	smbios {
```
samueldr, line 22:
```suggestion
+	smbios {
		compatible = "u-boot,sysinfo-smbios";
```

samueldr:
Hi!

Sorry it took this long to finally look into this. It's because I really needed to investigate whether we could provide the data implicitly to any Tow-Boot builds.

Turns out we can!

But, good news, this PR is still valid and good!!

The data should be part of the actual U-Boot source code, so this is a good example of how to do it *correctly* for a specific device.

I still have to test this patch with #158 applied, but I have strong reasons to believe this is all good.

samueldr:
Oh curse me... It should have targeted development and not released.

This is entirely my fault. (And GitHub's failure at providing good workflows...)

PR #145, opened 2022-04-16
[WIP] starfive-visionfive: init

Madouura:
I'll be honest, I have absolutely no idea what I'm doing and this is well outside my skill-set.
I think I can figure out the rest, but my main issue is I have no idea how to either generate or get a bin like I see in https://github.com/Tow-Boot/Tow-Boot/blob/released/modules/hardware/allwinner/default.nix#L77.
I think it might have something to do with this? https://github.com/starfive-tech/JH7100_secondBoot

### DIFF
--- a/modules/hardware/starfive/default.nix
+++ b/modules/hardware/starfive/default.nix
@@ -1,0 +1,68 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib)
+    mkIf
+    mkMerge
+    mkOption
+    types
+  ;
+
+  cfg = config.hardware.socs;
+  starFiveSOCs = [ "starfive-jh7100" ];
+  anyStarFive = lib.any (soc: config.hardware.socs.${soc}.enable) starFiveSOCs;
+in
+{
+  options = {
+    hardware.socs = {
+      starfive-jh7100.enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = "Enable when SoC is StarFive JH7100";
+        internal = true;
+      };
+    };
+  };
+
+  config = mkMerge [
+    {
+      hardware.socList = starFiveSOCs;
+    }
+    (mkIf anyStarFive {
+      system.system = "riscv64-linux";
+
+      Tow-Boot = {
+        defconfig = "starfive_jh7100_visionfive_smode_defconfig";
+
+        src = pkgs.fetchFromGitHub {
+          owner = "NickCao";
+          repo = "u-boot-starfive";
+          rev = "ac75aa54020412a83b61dad46c5ea15e7f9f525c";
+          sha256 = "1idh5k1479znp24rrfa0ikgk6iv5h80zscqhi6yv5ah4czia3ip3";
+        };
+
+        builder.additionalArguments = {
+          secondBoot = "${pkgs.buildPackages.Tow-Boot.jh7100-secondBoot}/${pkgs.buildPackages.Tow-Boot.jh7100-secondBoot.name}.bin";
+          ddrinit = "${pkgs.buildPackages.Tow-Boot.jh7100-ddrinit}/${pkgs.buildPackages.Tow-Boot.jh7100-ddrinit.name}.bin";
+        };
+
+        builder = {
+          installPhase = ''
+            ls -al
+          '';
+        };
+      };
+    })
+
+    # Documentation fragments
+    (mkIf (anyStarFive) {
+      documentation.sections.installationInstructions =
+        lib.mkDefault
+        (config.documentation.helpers.genericInstallationInstructionsTemplate {
+          # StarFive will prefer SD card always.
+          startupConflictNote = "";
+        })
+      ;
+    })
+  ];
+}
--- a/boards/starFive-visionFive/default.nix
+++ b/boards/starFive-visionFive/default.nix
@@ -1,0 +1,16 @@
+{
+  device = {
+    manufacturer = "StarFive";
+    name = "VisionFive";
+    identifier = "starFive-visionFive";
+    productPageURL = "https://starfivetech.com/en/site/exploit/";
+  };
+
+  hardware = {
+    soc = "starfive-jh7100";
+  };
+
+  Tow-Boot = {
+    defconfig = "starfive_jh7100_visionfive_smode_defconfig";
+  };
+}
--- a/modules/hardware/default.nix
+++ b/modules/hardware/default.nix
@@ -14,6 +14,7 @@ in
   imports = [
     ./allwinner
     ./amlogic
+    ./starfive
     ./rockchip
     ./generic.nix
   ];
--- a/modules/system.nix
+++ b/modules/system.nix
@@ -31,6 +31,7 @@ in
           "armv6l-linux"
           "armv7l-linux"
           "aarch64-linux"
+          "riscv64-linux"
         ];
         description = ''
           Defines the kind of target architecture system the device is.
--- a/modules/tow-boot/config.nix
+++ b/modules/tow-boot/config.nix
@@ -48,7 +48,6 @@ in
       BOOTDELAY = freeform "2";
 
       # 27 is ESCAPE
-      AUTOBOOT_MENUKEY = freeform "27";
 
       # So we'll fake that using CTRL+C is what we want...
       # It's only a side-effect.
@@ -63,7 +62,6 @@ in
       ;
 
       # And this ends up causing the menu to be used on ESCAPE (or CTRL+C)
-      AUTOBOOT_USE_MENUKEY = yes;
 
       # Additional commands
       CMD_BDI = yes;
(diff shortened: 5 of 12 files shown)


modules/hardware/starfive/default.nix
```diff
@@ -0,0 +1,50 @@
...
+    mkIf
+    mkMerge
+    mkOption
+    types
+  ;
+  inherit (config.hardware) mmcBootIndex;
+  cfg = config.hardware.socs;
+  starFiveSOCs = [ "starfive-jh7100" ];
+  anyStarFive = lib.any (soc: config.hardware.socs.${soc}.enable) starFiveSOCs;
+  isPhoneUX = config.Tow-Boot.phone-ux.enable;
```
samueldr: Until there are mobile platforms with starfive SoCs, we can drop any `phone-ux` configs.

modules/hardware/starfive/default.nix
```diff
@@ -0,0 +1,50 @@
...
+        # uhh...
+      };
+    })
+
+    # Documentation fragments
+    (mkIf (anyStarFive && !isPhoneUX) {
+      documentation.sections.installationInstructions =
+        lib.mkDefault
+        (config.documentation.helpers.genericInstallationInstructionsTemplate {
+          # StarFive will prefer SD card always.
```
samueldr, line 62: I don't know the platform. Is this true (that the platform prefers SD cards, always)?

samueldr (commented):
Some quick notes

samueldr:
Hi! :wave:

I don't know much about the StarFive platform at this moment, so I'm not sure how much I can help in figuring out the platform specific details. But, if you have the platform-specific knowledge, I can help apply it to a Tow-Boot build (as you've deduced already).

I'm not positive that the `SecondBoot` build will be needed for Tow-Boot. I would have to read more about the platform. It looks like they intend to have that `SecondBoot` manage starting the platform firmware, and also manage its upgrade. With what I've read, it really could go either way.

It looks like there is no information about the U-Boot "packaging" process in the `SeconBoot` repository you have linked, so to know more about the process for building the proper U-Boot binary, we would need to see how downstream (the vendor) or upstream (U-Boot) recommends doing it.

modules/hardware/starfive/default.nix
Madouura: For VisionFive, yes, I believe it always prefers SD cards.

modules/hardware/starfive/default.nix
Madouura: Ahh okay.

Madouura:
Currently, StarFive has u-boot here: https://github.com/starfive-tech/u-boot
It reads from the third partition of the sd card for uEnv.txt.
Specifics for VisionFive are here: https://github.com/starfive-tech/VisionFive
I honestly opened this a bit prematurely and should have done some more reading on the matter, but yeah I don't think secondBoot will be needed for tow-boot.
I should be getting my board today, so other than my own tests, if there's anything you would like to test, I'll be able to.

modules/hardware/starfive/default.nix
Madouura: Specifically, u-boot will always prefer (require) an sd card for booting.
I'm not entirely sure if in the firmware , that's a hard requirement.

Madouura: Specifically, u-boot will always prefer (require) an sd card for booting.
I'm not entirely sure if in the firmware , that's a hard requirement.

Madouura:
More info: https://github.com/starfive-tech/u-boot/blob/0c08d335c5acb6b1293f89d621c7271a27c20065/include/configs/starfive-jh7100.h#L58-L62
https://github.com/u-boot/u-boot/blob/c6ae38b38967a5c33d729c20e508a03ba3e0e3f6/include/configs/sifive-unmatched.h#L57-L70

samueldr:
Seeing the `.h` file for the board, it looks like they are definitely not using the generic distro boot command scheme from upstream U-Boot. Usage of uEnv in a partition also alludes to that.

This will need some additional work to fix. Having hardcoded partition schemes like this for the operating system goes against guidelines for standards-based boot.

Madouura:
Ah forgot about this.
image
From https://rvspace.org/en/Product/VisionFive/Technical_Documents/VisionFive_Single_Board_Computer_Quick_Start_Guide
So we probably only really need to replace the u-boot part with tow-boot.

Madouura:
And as a matter of fact, here we are: https://github.com/starfive-tech/Fedora_on_StarFive/releases
Now we just need to replace [fw_payload_visionfive.bin.out](https://github.com/starfive-tech/Fedora_on_StarFive/releases/download/fw_payload_jh7100_20220311/fw_payload_visionfive.bin.out) with tow-boot.

samueldr:
~~I haven't spent long looking, but is there documentation about where the SoC looks for the stages presented in your previous diagram?~~

The [ethernet programming section](https://rvspace.org/en/Product/VisionFive/Technical_Documents/VisionFive_Single_Board_Computer_Quick_Start_Guide#using-ethernet) gives enough information. Use of `sf probe` shows it's written to the SPI Flash, and the different `sf update` commands show how the flash is setup:

| Address | filename | What | Where |
|-------------|-----------|---------|--------|
| `0x0` | `bootloader-JH7100-211102.bin.out` | Bootloader [sic] / secondboot | https://github.com/starfive-tech/JH7100_secondBoot |
| `0x10000` | `ddrinit-2133-211102.bin.out` | RAM Training | https://github.com/starfive-tech/JH7100_ddrinit
| `0x40000` | `fw_payload_visionfive.bin.out` | "U-Boot" | https://github.com/starfive-tech/Fedora_on_StarFive/releases

Note the quotes around *U-Boot*. It is unclear what kind of involvement there is for the *OpenSBI* and *Linux* repositories [in the repository](https://github.com/starfive-tech/Fedora_on_StarFive). I suspect *Linux* is relevant only for the released Linux images. If I understand it right, OpenSBI is analogous to TF-A (ARM Trusted Firmware), so that will likely be required for U-Boot too.

* * *

We should strive to build every stage here from source. This is because it will more closely match the usual setup where the Tow-Boot image includes *everything required* for itself. This way other assumptions hold better, e.g. we know that we can *just* splat the image to the whole SPI Flash.

I'll also note that [there are instructions to recover for emptied / bad SPI flashes](https://rvspace.org/en/Product/VisionFive/Technical_Documents/VisionFive_Single_Board_Computer_Quick_Start_Guide#appendix-c-recovering-the-bootloader). It looks sane enough to rely on for development purposes, and *not too bad* for end-users in the rare off-chance of a mishap.

modules/hardware/starfive/default.nix
samueldr: Right, I understand the confusion.

Generally speaking, there are two "orders" when you power on a system.

The *Platform Firmware startup order*, and the *operating system boot order*. I make it a point to reserve the word "boot" to operating systems only.

This comment is about the *Platform Firmware startup order*, sometimes called *BootROM order*. This is the locations the SoC will look for U-Boot, and start it from. I don't think the BootROM even looks at other locations than SPI Flash [and serial], but I have literally no proof about that. So this comment will not hold true.

Note that the *operating system boot order* is entirely controlled by the platform firmware (U-Boot, Tow-Boot), and with Tow-Boot we set it to what we want, which is "internal targets first".

Madouura: I'm not entirely sure now, if we can replace the boot stuff in the SPI, we might be able to boot from USB.

samueldr: I am not 100% sure, but given what I've read yesterday, I'm about 90% sure that the firmware will strictly be loaded from SPI Flash, and that the OS will be loaded *however we want*.

And that indeed means that if USB support is complete enough in U-Boot for the board, that it will be possible to boot from a good old boring and generic USB UEFI iso.

Madouura:
Okay, I see how we build u-boot now.
Looks like we may need to override the Two-Boot src attribute to use starfive's repo.

Madouura:
Or use github's compare HASH..HASH feature with fetchpatch, but I'm unsure about either option for this.
Would it be better to wait until this is in upstream u-boot?
Especially considering starfive's u-boot updates semi-frequently.

samueldr:
You probably would want to keep this PR open as a WIP that is updated regularly, allowing you and others to already benefit from Tow-Boot, and get more testing done before it's upstreamed.

Either way would work, but there might be issues applying on an arbitrary U-Boot version, as Tow-Boot patches might not all apply.

Madouura:
I'll see what I can do with it in this PR, while we wait for upstream.

Madouura:
So, Tow-Boot already combines everything together, correct?
Because, the normal process requires combining u-boot into opensbi, and that's something I'm trying to do here, but I'm unsure if that _should_ be done.
Is this perhaps easier than I think and we can just let Tow-Boot combine everything together?

PR #146, opened 2022-04-16
doc/getting-started: ask to grow the GPT partition to physical size

### LINKED ISSUE #52: Shared Storage guidance (fdisk warns about GPT?)
Hi.

Maybe this is because my SD came pre-formatted, but after I `dd` the rpi4 `shared.disk-image.img` to my SD Card... I figured I'd fire up `fdisk` to add a traditional `/boot` and `/root` for a shared media install.

However, fdisk gave me this warning, and I'm not actually sure what I want to do here:

```
/home/cole/downloads/raspberryPi-aarch64-2021.04-002> sudo dd if=shared.disk-image.img of=/dev/sdb bs=4M; sudo sync; sync
8+1 records in
8+1 records out
34620416 bytes (35 MB, 33 MiB) copied, 0.018304 s, 1.9 GB/s
```

```
/home/cole/downloads/raspberryPi-aarch64-2021.04-002> sudo fdisk /dev/sdb

Welcome to fdisk (util-linux 2.36.2).
Changes will remain in memory only, until you decide to write them.
Be careful before using the write command.

The backup GPT table is not on the end of the device. This problem will be corrected by write.
A hybrid GPT was detected. You have to sync the hybrid MBR manually (expert command 'M').
```

Any tips? Should I ignore it? Should I have "wiped" the sd card more thoroughly?

flokli:
Otherwise, fdisk and friends first warn about the GPT being possibly
"corrupt", because the backup GPT table is not at the end of the medium.

Fixes https://github.com/Tow-Boot/Tow-Boot/issues/52.

### DIFF
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -85,6 +85,15 @@ but the size has been chosen to allow further expansion if needed.
    ```
     # dd if=shared.disk-image.img of=/dev/XXX bs=1M oflag=direct,sync status=progress
    ```
+
+1. Some partitioning tools require the secondary GPT to be at the end of the
+   disk before attempting any operations. Grow the GPT partition to the physical
+   size of the medium. (This operation is a no-op that forces a rewrite of the
+   partition table.)
+   ```
+    # sfdisk --append /dev/XXX
+   ```
+
 1. Boot the media on your board
 
 > **NOTE**: When installing your operating system, make sure that you do not
--- a/doc/getting-started.md
+++ b/doc/getting-started.md
@@ -89,8 +89,8 @@ but the size has been chosen to allow further expansion if needed.
 
 > **NOTE**: When installing your operating system, make sure that you do not
 > re-create the partition table from scratch, only edit the existing one.
-> The *Installing an Operating System* section from this manual will remind
-> you.
+> The [*Installing an Operating System*](installing-an-operating-system.md)
+> section from this manual will remind you.
 
 Using the *shared storage* strategy, you can simulate *dedicated storage* by
 **not** installing and using the storage media the platform firmware lives


samueldr:
Hi,

Sorry it's taking this long until I look correctly into this.

There is this odd dichotomy with the project where it needs to be *technically accurate*, but also tries to reduce any appearance of jank. I have seen some FUD thrown around already due to how Tow-Boot transgresses the status quo. Requiring "this weird step" after writing the image could feel like jank when compared to the status quo of disk image distributions.

So first, what does the GPT spec (part of UEFI) states about this situation? Nothing *exactly* applies. The closest is the following paragraph, boldened emphasis by me:

> Both the primary and backup GPTs must be valid before an attempt is made to grow the size of a physical volume. This is due to the GPT recovery scheme depending on locating the backup GPT at the end of the device. A volume may grow in size when disks are added to a RAID device. **As soon as the volume size is increased the backup GPT must be moved to the end of the volume and the primary and backup GPT Headers must be updated to reflect the new volume size.**

Interpreting this quite strictly, yes, instructions to fix the GPT *should* be part of these instructions.

Though, there is a problem. `sgdisk` is relatively non-standard. It's part of `gptfdisk`, which I wouldn't expect to necessarily be part of a Linux "base system".

The more standard (in `util-linux`) `sfdisk` can fix it too:

```
~ $ sudo sfdisk --verify /dev/mmcblk0
GPT PMBR size mismatch (26657 != 3887103) will be corrected by write.
The backup GPT table is not on the end of the device.
/dev/mmcblk0:
MyLBA mismatch with real position at backup header.
1 error detected.

~ $ sudo sfdisk --relocate gpt-bak-std /dev/mmcblk0
GPT PMBR size mismatch (26657 != 3887103) will be corrected by write.
The backup GPT table is not on the end of the device. This problem will be corrected by write.

The partition table has been altered.
Calling ioctl() to re-read partition table.
Syncing disks.

~ $ sudo sfdisk --verify /dev/mmcblk0
/dev/mmcblk0:
No errors detected.
Header version: 1.0
Using 1 out of 128 partitions.
A total of 3862461 free sectors is available in 2 segments (the largest is 1.8 GiB).
```

Any reasons not to go with `sfdisk` you think?

I would hazard a guess that since [it's from 2.36, (mid 2020)](https://github.com/util-linux/util-linux/blob/441f9b9303d015f1777aec7168807d58feacca31/Documentation/releases/v2.36-ReleaseNotes#L44-L46) the feature could be missing from some larger distros with longer release cycles.

* * *

*(You don't need to respond to this part, I'm only venting a bit about the spec and implementations...)*

With that said, I believe this is an oversight in the spec. There is nothing specifying what should happen with partitioning software in that situation. Nothing specifies what partitioning software should do when the backup GPT is *where it is stated it should be in the header*, but is not at the end of the disk.

Considering the previous quote stating *as soon as the volume size is increased*, I would argue that it is the partitioning software's duty to apply this transformation if not handled by whatever process "caused the expansion". After all, in this situation the GPT is entirely valid and coherent with itself, it's just a bit out of spec. The spec already has passages specifying what should happen in worse situations, and basically amounts to "the partitioning software must fix it when possible".

samueldr:
With previous `sfdisk` releases, it's possible to do so by forcing a no-op write operation this way:

```
 $ NIX_PATH=nixpkgs=channel:nixos-20.03 nix-shell -p utillinux
...

[nix-shell:~]$ sfdisk --version
sfdisk from util-linux 2.33.2

[nix-shell:~]$ sudo sfdisk --verify  /dev/mmcblk0
GPT PMBR size mismatch (26657 != 3887103) will be corrected by write.
The backup GPT table is not on the end of the device. This problem will be corrected by write.
/dev/mmcblk0:
No errors detected.
Header version: 1.0
Using 1 out of 128 partitions.
A total of 3862461 free sectors is available in 2 segments (the largest is 1.9 GiB).

[nix-shell:~]$ echo | sudo sfdisk --append  /dev/mmcblk0
GPT PMBR size mismatch (26657 != 3887103) will be corrected by write.
The backup GPT table is not on the end of the device. This problem will be corrected by write.
Checking that no-one is using this disk right now ... OK

Disk /dev/mmcblk0: 1.9 GiB, 1990197248 bytes, 3887104 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: gpt
Disk identifier: E0CA6E57-39B2-4482-9838-21E2785CD93D

Old situation:

Device         Start   End Sectors Size Type
/dev/mmcblk0p1    64 24639   24576  12M unknown

/dev/mmcblk0p2: Done.

New situation:
Disklabel type: gpt
Disk identifier: E0CA6E57-39B2-4482-9838-21E2785CD93D

Device         Start   End Sectors Size Type
/dev/mmcblk0p1    64 24639   24576  12M unknown

The partition table has been altered.
Calling ioctl() to re-read partition table.
Syncing disks.

[nix-shell:~]$ sudo sfdisk --verify  /dev/mmcblk0
/dev/mmcblk0:
No errors detected.
Header version: 1.0
Using 1 out of 128 partitions.
A total of 3862461 free sectors is available in 2 segments (the largest is 1.9 GiB).
```

flokli:
@samueldr I don't have a particular preference to use `sgdisk` over `sfdisk`.

Can you update this with the right `sfdisk` invocation and merge it?

samueldr:
Sure, it was less about the specific change, but to ask and see if I had missed anything :).

PR #148, opened 2022-04-22
pine64-pinephonePro: Use appropriate OPP

samueldr:
The vendor requested that the maximum frequencies and maximum voltage
respect the RK3399-T datasheet.

This applies to all Pinephone Pro, even those without an RK3399-T, be it
the unseen RK3399s, or the standard RK3399. This is because while the
SoC may handle it fine, the design was validated with the lower
frequencies, and lower thermals. The LCD panel may operate out of range
when the device is clocked higher than designed for.

See also:

 - https://gitlab.com/pine64-org/linux/-/merge_requests/36

* * *

### What was done

Update the SPI-installed Tow-Boot with this PR's changes. Still works as expected.

* * *

Note that in practice, this won't be a change that is held only in Tow-Boot. In fact this patched change should disappear and coalesce in a totally innocuous "sync" of the kernel DTs.

### DIFF
--- a/boards/pine64-pinephonePro/0001-pine64-pinephonepro-device-enablement.patch
+++ b/boards/pine64-pinephonePro/0001-pine64-pinephonepro-device-enablement.patch
@@ -2647,3 +2647,154 @@ index c51d1657a2a..22c2ced2d79 100644
 -- 
 2.34.0
 
+From 132da837a96cb3cfbbf2937d3e9e743d7005b548 Mon Sep 17 00:00:00 2001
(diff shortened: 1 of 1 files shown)


samueldr:
There are still unanswered questions about the linux-side PR, so let's push it to the next release.

samueldr:
It's been handled differently in Linux.

Instead of a new OPP list, they are edited in the Pinephone Pro dts. I guess it's expected there will never be another RK3399-T/S/whichever added to Linux?

PR #149, opened 2022-04-26
odroid-N2: Add support for XT25Q64D

ulrikstrid:
This adds a build for the odroid N2+ computer. Also adds support for the XTX SPI XT25Q64D

### DIFF
--- a/boards/odroid-N2/0002-Add-support-for-XTX-SPI-XT25Q64D.patch
+++ b/boards/odroid-N2/0002-Add-support-for-XTX-SPI-XT25Q64D.patch
@@ -1,0 +1,35 @@
+From ba58f0f3eadc1755158dec3ea1c063d53c0e702a Mon Sep 17 00:00:00 2001
+From: Ulrik Strid <ulrik.strid@outlook.com>
+Date: Tue, 26 Apr 2022 09:27:57 +0200
+Subject: [PATCH 2/2] Add support for XTX SPI XT25Q64D
+
+Signed-off-by: Ulrik Strid <ulrik.strid@outlook.com>
+---
+ configs/odroid-n2_defconfig   | 1 +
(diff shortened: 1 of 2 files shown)


samueldr (changes requested):
:wave: Hi!

Thank you for the contribution!

* * *

We can do more simpler. The N2 configuration is already used for N2+.

In fact, your problem is not plus vs. non-plussed N2, but that some of the batches (new ones?)  apparently are using a new yet-unsupported SPI Flash.

So, really, you can condense this into only:

> Add support for XTX SPI XT25Q64D

And apply it to the existing N2 configuration.

Adding a note here that this is desirable. When possible, board configurations should be coalesced into a single one.

samueldr:
I'll need to build it for verification, but I don't have the hardware to test. The changes look okay. You probably want to rename the PR something like "odroid-N2: Add support for XT25Q64D"

I assume that you have now been able to successfully flash to SPI, and start your N2+ from SPI, right?

ulrikstrid:
I have not tested it with the new changes, but it builds, I can try it tomorrow.
The initial state of this PR flashes and I have managed to boot NixOS on top

ulrikstrid:
Note that the patch will not apply cleanly in upstream u-boot since they don't have the first patch.

samueldr:
> Note that the patch will not apply cleanly in upstream u-boot since they don't have the first patch.

Either you'll have submitted something to upstream, or we rebase things ourselves, so it's not a big deal. Thanks for the heads-up.

ulrikstrid:
@samueldr I can confirm that I can flash the SPI and boot into a OS with the current state of this PR

samueldr:
(Changed the base branch to force a refresh of the PR... and yesterday I though the "refresh commits" option in SourceForge was weird...)

samueldr (approved):

samueldr:
Thank you!

PR #160, opened 2022-06-29
Add support for the ODROID-C4

tpwrules:
Verified that the following functionality works:
* HDMI output and USB keyboard input
* Serial console input and output using ODROID's official dongle and connector
* Extlinux boot off SD, eMMC, and USB devices of official NixOS 22.05 kernel 5.15 SD images
* DHCP obtainment of IP address

Unfortunately UEFI boot does not work properly, likely due to U-Boot shipping an incompatible device tree.

I tested installing NixOS by adding a root partition using `fdisk` and making it bootable, then `dd`ing the root partition from the official SD image (not the whole image) over to my new partition. I also tested that it's possible to `dd` the whole SD image to a flash drive and boot it, but that's kind of silly.

I'm not sure if the Amlogic S905X3 is deserving of its own SM1 family group as the boot logic and preparation is evidently compatible with the G12 family. For now I decided to just place it with the G12 logic.

This work should also be compatible with the ODROID-HC4 through little more than adding an `h` to the board and device tree name. But I don't have one to test with.

### DIFF
--- a/modules/hardware/amlogic/default.nix
+++ b/modules/hardware/amlogic/default.nix
@@ -18,6 +18,7 @@ let
   amlogicG12 = lib.any (soc: config.hardware.socs.${soc}.enable) [
     "amlogic-a311d"
     "amlogic-s922x"
+    "amlogic-s905x3" # technically an SM1 family member, but the boot process is identical to G12
   ];
   amlogicGXL = lib.any (soc: config.hardware.socs.${soc}.enable) [
     "amlogic-s805x"
@@ -54,6 +55,12 @@ in
         description = "Enable when SoC is Amlogic S905";
         internal = true;
       };
+      amlogic-s905x3.enable = mkOption {
+        type = types.bool;
+        default = false;
+        description = "Enable when SoC is Amlogic S905X3";
+        internal = true;
+      };
       amlogic-s922x.enable = mkOption {
         type = types.bool;
         default = false;
@@ -69,6 +76,7 @@ in
         "amlogic-a311d"
         "amlogic-s805x"
         "amlogic-s905"
+        "amlogic-s905x3"
         "amlogic-s922x"
       ];
     }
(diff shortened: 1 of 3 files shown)


samueldr:
> Unfortunately UEFI boot does not work properly, likely due to U-Boot shipping an incompatible device tree.

Putting the "correct" (kernel-built) dtb files under `ESP:/dtb/` would allow verifying the assumption.

modules/hardware/amlogic/default.nix
```diff
@@ -18,6 +18,7 @@ let
   amlogicG12 = lib.any (soc: config.hardware.socs.${soc}.enable) [
     "amlogic-a311d"
     "amlogic-s922x"
+    "amlogic-s905x3" # technically an SM1 family member, but the boot process is identical to G12
```
samueldr, line 21: It's correct to do so at this point *since you documented it correctly*.

tpwrules:
I can confirm that loading a kernel-built device tree with grub fixes the SD card not being detected during UEFI boot.

tpwrules:
If U-Boot loads the device tree then that also fixes the issue. But it still doesn't always detect my flash drive and when that happens, the stage 1 error prompt ends up on the serial console so you have to hook that up to manually mount the drive. I'm not sure how to fix that. There are some kernel messages on the HDMI display but it's not the default console.

These same problems happen on my ODROID-N2, so maybe my flash drive is weird and NixOS's root detection timeout needs to be extended a bit. It would also be great to fix the console situation, but I don't know if that can go in two places at once and I don't know how many users use HDMI and KB/mouse vs. serial.

samueldr:
[I know you kinda solved it over in the NixOS matrix room, but here for posterity]

Kernel messages are output by default on `/chose/stdout=path`, or else on `console=` parameters. But note that `init` will only write to the *last valid `console=` parameter*, unless there is some form of multiplexing available.

So in this case, to output to the VT, you would use `console=tty0` (`tty0` outputs to any currently visible VT), or to `ttyAML0` or something like that for serial *on amlogic*.

PR #162, opened 2022-07-01
Add tips about boot errors

Bujiraso:
First PR here, open to any edits or suggestions!

I'm looking to fill out what I see as a missing piece of valuable intel on the FAQ. Thanks!


Bujiraso:
Hm, actually -- re-installing Tow-boot *WAS* necessary to fix it. I'll close this as I'm not sure it's correct.

PR #163, opened 2022-07-19
bananaPi-p2Maker: Add support for Banana Pi P2 Maker

leo60228:
The Banana Pi P2 Maker is a low-cost Allwinner H2+ (variant of the H3) development board primarily notable for having (optional) PoE. I've tested this with Arch Linux ARM, Debian, and Fedora, and all three work. It likely also would work with the P2 Zero, which uses the same PCB but with eMMC (and possibly support components for it?) populated, but of course without eMMC support. I initially attempted implementing the P2 Zero instead of the P2 Maker, but I wouldn't have been able to test the eMMC capabilities, and the missing eMMC caused warnings during boot.

The U-Boot patch could probably be upstreamed (especially the change to `sunxi-h3-h5.dtsi`, which is necessary for USB host functionality to work in U-Boot after a refactor), but I've had bad experiences submitting patches to these projects.

### DIFF
--- a/boards/bananaPi-p2Maker/0001-sunxi-add-Banana-Pi-P2-Maker.patch
+++ b/boards/bananaPi-p2Maker/0001-sunxi-add-Banana-Pi-P2-Maker.patch
@@ -1,0 +1,359 @@
+From 68d732469510d989e12079069665ba247bf3f7c4 Mon Sep 17 00:00:00 2001
+From: leo60228 <leo@60228.dev>
+Date: Fri, 15 Jul 2022 15:25:17 -0400
+Subject: [PATCH] sunxi: add Banana Pi P2 Maker
+
+The Banana Pi P2 Maker is closely related to the already supported Banana
+Pi M2 Zero, but adds Ethernet with optional PoE.
+---
+ arch/arm/dts/Makefile                         |   1 +
+ .../dts/sun8i-h2-plus-bananapi-p2-maker.dts   | 281 ++++++++++++++++++
+ arch/arm/dts/sunxi-h3-h5.dtsi                 |   4 +
+ configs/bananapi_p2_maker_defconfig           |  13 +
+ 4 files changed, 299 insertions(+)
+ create mode 100644 arch/arm/dts/sun8i-h2-plus-bananapi-p2-maker.dts
+ create mode 100644 configs/bananapi_p2_maker_defconfig
+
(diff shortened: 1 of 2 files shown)


samueldr:
Hi! :wave:

Given this is an Allwinner board, the changes are 99.99% likely good without requiring any more testing. In my experience they all work so much alike that it's (thankfully) boring.

I don't see you specifically stating that you've tested HDMI out. Have you?

Otherwise it all looks good. I don't really see any reason not to merge this.

* * *

> especially the change to `sunxi-h3-h5.dtsi`, which is necessary for USB host functionality to work in U-Boot after a refactor

I don't think I'll have the time to test, but I wonder if any h3/h5 boards I own have been affected. I guess the way to check is that USB plainly doesn't work? Or is it a particular interface? I'm led to understand that the OTG port at the very least would be affected.

* * *

> The U-Boot patch could probably be upstreamed

It really should, I've not been stellar at it, but ideally only opinionated changes should live in this repo. Or at least I'd like that every changes sees a genuine effort at being sent upstream.

> but I've had bad experiences submitting patches to these projects.

Assuming it's about the individuals, have you had experience with U-Boot specifically? You can leave the answer vague if you're not comfortable, and I definitely don't want names in that unlikely event.

I've found that the worst that happened to me was that a set of patches was left ignored, though asking around on their IRC channel gives an almost direct link to some maintainers that can help. My only complaint is that it's the ol' patch-in-mail, everything else has been good.

leo60228:
> I don't see you specifically stating that you've tested HDMI out. Have you?

Yes. There's some weird artifacting at 4K without accelerated graphics, but it's still usable for setup. 

> I guess the way to check is that USB plainly doesn't work? Or is it a particular interface? I'm led to understand that the OTG port at the very least would be affected.

I *think* it's just the OTG port, and only when device mode support is compiled out in U-Boot, but this device only has the one port and I didn't do much testing with device mode enabled.

> Assuming it's about the individuals, have you had experience with U-Boot specifically? You can leave the answer vague if you're not comfortable, and I definitely don't want names in that unlikely event.

It's not really about any individuals.

samueldr:
Oops... Got in the flow with other PRs and rebased+pushed, even though this applies as-it-is on `development`. The push is a rebase without changes.

Can you verify this still works fine on the device with 2022.07? I've built it and it built without issues, without any change.

Given how boring this update cycle (2021.10 → 2022.07) was for Allwinner, it wouldn't surprise me things just work.

PR #165, opened 2022-07-27
pine64-pinecube: init board

lheckemann:
This works, but still has a number of problems:
 - can't boot kernels bigger than 8MiB
 - initramfs doesn't seem to work either (may also be a size issue)
 - extra patches[1] might be necessary for ethernet to work

[1]: https://github.com/danielfullmer/pinecube-nixos/blob/master/uboot/Pine64-PineCube-uboot-support.patch

### DIFF
--- a/boards/pine64-pinecube/default.nix
+++ b/boards/pine64-pinecube/default.nix
@@ -8,10 +8,14 @@
 
   hardware = {
     soc = "allwinner-s3";
+    SPISize = 16 * 1024 * 1024; # 128mbit
   };
 
   Tow-Boot = {
     defconfig = "pinecube_defconfig";
     withLogo = false;
+    patches = [
+      ./0001-pinecube-enable-SPI-booting-flashing.patch
+    ];
   };
 }
--- a/boards/pine64-pinecube/default.nix
+++ b/boards/pine64-pinecube/default.nix
@@ -1,0 +1,17 @@
+{
+  device = {
+    manufacturer = "PINE64";
+    name = "PineCube";
+    identifier = "pine64-pinecube";
+    productPageURL = "https://www.pine64.org/cube/";
+  };
+
+  hardware = {
+    soc = "allwinner-s3";
+  };
+
+  Tow-Boot = {
+    defconfig = "pinecube_defconfig";
+    withLogo = false;
+  };
+}
--- a/modules/hardware/allwinner/default.nix
+++ b/modules/hardware/allwinner/default.nix
@@ -13,6 +13,7 @@ let
     "allwinner-a64"
     "allwinner-h3"
     "allwinner-h5"
+    "allwinner-s3"
   ];
   anyAllwinner = lib.any (soc: config.hardware.socs.${soc}.enable) allwinnerSOCs;
   anyAllwinner64 = anyAllwinner && config.system.system == "aarch64-linux";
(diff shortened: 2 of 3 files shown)


modules/hardware/allwinner/default.nix
```diff
@@ -94,14 +101,17 @@ in
     (mkIf cfg.allwinner-a64.enable {
       system.system = "aarch64-linux";
     })
-    (mkIf cfg.allwinner-h3.enable {
+    (mkIf (cfg.allwinner-h3.enable) {
```
samueldr: I guess this is an accidental needless change from initially adding `&& cfg.allwinner-s3.enable`?

Anyway, that's not needed.

```suggestion
    (mkIf cfg.allwinner-h3.enable {
```

modules/hardware/allwinner/default.nix
```diff
@@ -94,14 +101,17 @@ in
...
       system.system = "armv7l-linux";
       Tow-Boot.config = [
         (helpers: with helpers; {
           CMD_POWEROFF = no;
         })
       ];
     })
+    (mkIf (cfg.allwinner-s3.enable) {
```
samueldr:
```suggestion
    (mkIf cfg.allwinner-s3.enable {
```

Not that it changes the result, but to keep the style consistent here.

* * *

Additionally, this should be placed after the `allwinner-h5` section, since this is ordered by name. (Not that it was documented to be,)

boards/pine64-pinecube/default.nix
```diff
@@ -0,0 +1,17 @@
...
+  device = {
+    manufacturer = "PINE64";
+    name = "PineCube";
+    identifier = "pine64-pinecube";
+    productPageURL = "https://www.pine64.org/cube/";
+  };
+
+  hardware = {
+    soc = "allwinner-s3";
+  };
```
samueldr, line 12: I believe simply adding the SPI setting here would be enough to make it a valid build. Allwinner targets are extremely simple to support.

samueldr (dismissed):
Some minor changes. Nothing that will change the overall result.

samueldr:
> This works, but still has a number of problems:
> 
> * can't boot kernels bigger than 8MiB
> * initramfs doesn't seem to work either (may also be a size issue)
> * extra patches[1](https://github.com/danielfullmer/pinecube-nixos/blob/master/uboot/Pine64-PineCube-uboot-support.patch) might be necessary for ethernet to work

I figure all of these would be upstream issues, right?

modules/hardware/allwinner/default.nix
lheckemann: Excellent sleuthing.

boards/pine64-pinecube/default.nix
lheckemann: I'm reluctant to add this without testing, and testing it is not quite at the top of my priority list. Eventually though :)

samueldr: Good mindset. I'll dig in my box of projects and get mine out to test it.

Note that on Allwinner it's also basically impossible to "brick", since it will always prefer starting from SD. So testing the installer should be a trivial matter.

samueldr:
This force push

 - Synced with current `development` branch
 - Added SPI support

I verified SPI support works, by using it on my board.

samueldr:
Before merging I'd like to verify and fix (from the PR description)

> - can't boot kernels bigger than 8MiB
> - initramfs doesn't seem to work either (may also be a size issue)

If this is an unsolveable limitation of the platform, it should be documented as such.

lheckemann:
I don't think it's an unsolveable limitation of the platform, just of the memory layout used by the default config.

samueldr:
Yeah, this was *if* it is... I'm positive it must be something like that too.

PR #168, opened 2022-08-02
Pinephone Pro: Add 2A current limit to PMIC to fix low battery bootloop

ghost:
Hello!!

This is a tiny patch based on the work megi has done in their levinboot firmware. By setting the register 0xA1, 'USB_CTRL_REG', to 0x77 it sets the current limit to 2A and therefore allowing the device take a good amount of charge before the OS has started. Moreover, the recent commits on their kernel fork has allowed for the set current limit at boot stay until the OS can change to a more sutible value. 

This should iradicate the bootloop issue completely. In their levinboot it works fine so it should work fine here too.

I haven't had chance to test it but theoretically it should work, i'll update this PR when i get round to testing it tomorrow unless someone beats me to it :)

All the best,

Joe.

### DIFF
--- a/boards/pine64-pinephonePro/0001-Set-RK8xx-current-limit-to-2A.patch
+++ b/boards/pine64-pinephonePro/0001-Set-RK8xx-current-limit-to-2A.patch
@@ -1,0 +1,27 @@
+From 16bc3396de37a0578252a70303892006085ece1d Mon Sep 17 00:00:00 2001
+From: Joe Todd <j.t0dd@protonmail.com>
+Date: Tue, 2 Aug 2022 18:03:45 +0100
+Subject: [PATCH] Set RK8xx current limit to 2A
+
+---
+ drivers/power/pmic/rk8xx.c | 2 ++
+ include/power/rk8xx_pmic.h | 2 +-
+ 2 files changed, 3 insertions(+), 1 deletion(-)
+
+diff --git a/drivers/power/pmic/rk8xx.c b/drivers/power/pmic/rk8xx.c
+index 25ef621f8d..d348d3f40d 100644
+--- a/drivers/power/pmic/rk8xx.c
++++ b/drivers/power/pmic/rk8xx.c
+@@ -200,6 +200,8 @@ static int rk8xx_probe(struct udevice *dev)
+ 	case RK818_ID:
+ 		on_source = RK8XX_ON_SOURCE;
+ 		off_source = RK8XX_OFF_SOURCE;
++
++		pmic_reg_write(dev, REG_USB_CTRL, 0x77); /* 0x77 gives current limit of 2A */ 
+ 		break;
+ 	case RK809_ID:
+ 	case RK817_ID:
+ 
+-- 
+2.37.1
+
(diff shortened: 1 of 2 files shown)


samueldr:
Hi!

Thanks for the contribution.

Currently, the way this PR is made, the patch isn't being added to the build. The patch would need to be added [to the `patches` list here](https://github.com/Tow-Boot/Tow-Boot/blob/b955c52c875d2ed155169c0bcb56ec7516dddb08/boards/pine64-pinephonePro/default.nix#L52-L78). Added last is likely fine.

Additionally, though it will require some git knowledge, it would be nice if you could rebase and squash the "edit" commit into a single commit, and use the form `board-name: message` as the commit title. If it's something you're not comfortable doing, no worries, I can handle it later on.

(I'll properly look at the change later.)

ghost:
Hi there!

Thank you for your feedback, this is my first time every doing anything with Tow-boot so forgive me. I can do all that tomorrow sometime unless you beat me to it.

Thanks again,

Joe.

ghost:
Hope that is a little bit better, sorry for the delay

samueldr:
> Hope that is a little bit better, sorry for the delay

Haha, don't be. You're not "on the hook" for anything here. The change set looks okay.

Can you described how (and if) you tested and validated the change?

No worries if in the end you haven't tested, I want to know so that if results diverge on my end I can make a guess for the reasons why.

ghost:
I am just about to try and figure out how to build Tow-Boot with this patch, i'm also trying to drain my pinephone pro as we speak. I'll report back asap :)

ghost:
Some notes from testing: Last night when the battery was completely flat (the phone didn't show any signs of life at all), it took two tries to boot Linux. The phone started the first time and crashed and then the next time it started and booted fine. 

This morning however, the battery again is completely flat and plugging it into a power source allows the phone to just boots straight up. Would like some more testing on this done, i'll do some today with older kernels and test again with Tow-Boot without this patch.

ghost:
Update: I moved the latest released Tow-Boot version without this patch onto the SPI and the phone bootloops continuesly when trying to boot an OS. Moving back to Tow-Boot with this patch and using megis kernel it boots first time, with stock it doesn't. This is the patch for the kernel that is like part 2 of fixing the boot loop issue: https://xff.cz/git/linux/commit/?h=ppp-drivers-5.19&id=bd62b82172bab9298fa9f3a337a6a817435ebfae

ghost:
Hi there!

Just wanted to give a small update on testing. With this patch on the SPI, using a fresh flash of the latest manjaro plasma mobile dev to the SD card and a completely flat battery the phone boots first time without issue.

khimaros:
is there an equivalent solution possible for pinephoneA64?

samueldr:
Hi :wave:, sorry it took this long until I could finally review and look into this.

We probably should go with the following instead:

 - https://xff.cz/git/u-boot/commit/?h=ppp&id=7f8238fd608290152b143322178a5be21a447dc1

One difference with that one is it uses 1.5A, but the implementation seems more proper. I also hope that this other implementation will/can be upstreamed. 

This current PR is serviceable to help alleviate issues, but is not upstreamable. (Though I think you knew.)

samueldr:
Even though it was superseded, thank you for the contribution!

PR #188, opened 2022-09-02
Podman script addition

psstoyanov:
Basic Podman helpers:

- [x] Check if base NixOS image is present
- [x] Check if Volume is present

Can be executed in rootless mode following the Podman docs: https://podman.io/getting-started/installation#linux-distributions

No changes made to the existing Docker script


psstoyanov:
Cleaning up afterwards will be nice but I'm not sure how to do that yet (or re-using the existing container/volume)

samueldr:
> Cleaning up afterwards will be nice but I'm not sure how to do that yet (or re-using the existing container/volume)

We probably don't want to clean-up, at least not the Nix store volume, as it reduces build time for further invocations. Though it might be relevant to document how to clean it up?

Can someone who knows podman also review and chime in? Thanks!

PureTryOut:
What's your reason for making an entirely different script rather than just changing the 2 Docker commands in the existing script? You can just detect if Podman is installed and if so prefer that one, including the extra command to check for existing volumes.

psstoyanov:
@PureTryOut , separating the two at least for now seems like the safest approach. I don't want to break something that is already working.
I haven't setup Docker to check if it would work with those changes either.

I don't know how readable it would be to hide what is used behind another variable (or how that var would be named). Although it might have the benefit of passing info to the user early as of what can be used :thinking: 

Note:
I haven't tried to pass which command to be used for other commands in `sh` script before. 

I'm sure there is more that can be improved but it's a start.

psstoyanov:
Alright, complete revamp of the previous iteration, @samueldr 

@PureTryOut - now both `docker` and `podman` are used within the same script, I hope this will be satisfactory and easy to read.

Added:
- [x] Check if the script has been executed as `root` user
- [x] Selection between `docker` && `podman`
- [x] Same image && volume commands for `docker`
- [x] Errors for the selection
- [x] Call `build.sh` script

Things which I'm uncertain if I implemented correctly:
- [ ] Exit codes
- [ ] Style of promps

Other:
- Feedback from other systems (MacOS?) will be good
- I'm not using `docker` much so another look there from other users is preferable 

Nice to have but most likely for a different PR:
- [ ] Adding board selection based on the dirs in [boards](https://github.com/Tow-Boot/Tow-Boot/tree/released/boards)?

support/docker/build.sh
```diff
@@ -29,11 +29,136 @@ stderr.printf() {
...
 }
 
-_nix_docker() {
+
+_nix_container_manager() {
+
+	# Ask if the user wants to proceed as root
+	if (($(id -u) == 0)); 
+	then
+		stderr.printf "\nYou don't need to run this script as root.\n\nLear more on how run docker/podman in rootless mode:\n Docker: https://docs.docker.com/engine/security/rootless/\n Podman: https://github.com/containers/podman/blob/main/docs/tutorials/rootless_tutorial.md\n\n"
```
samueldr:
```suggestion
		stderr.printf ""
		stderr.printf "You don't need to run this script as root."
		stderr.printf ""
		stderr.printf "Learn more on how run docker/podman in rootless mode:"
		stderr.printf " Docker: https://docs.docker.com/engine/security/rootless/"
		stderr.printf " Podman: https://github.com/containers/podman/blob/main/docs/tutorials/rootless_tutorial.md"
		stderr.printf ""
```

Long spans of text without newlines are harder to read.

Note how `Lear` was changed to `Learn`; easier to spot when reading :)

This is not high performance code, putting multiple printfs one after the other is more correct since it helps the developer more.

support/docker/build.sh
```diff
@@ -29,11 +29,136 @@ stderr.printf() {
...
+				No)
+					stderr.printf "\nQuiting the build script...\n"
+					exit 0
+					;;
+			esac
+		done
+	fi
+
+	# TODO: Add selection menu of available boards located in boards dir
+
```
samueldr:
```suggestion
```

Let's not add TODOs about changing the form of this script.

This script should provide the same semantics as `nix-build` does, since it's meant to allow non-Nix users to get acclimated into building with `nix-build`.

If we want a non-nix-build CLI helper to help people select what to build, it should be done externally to that script, and in a way that also works with `nix-build` without the container stuff. (note: I *do not know* if I want that, and that would be more for external potential contributors to discuss in a new issue/PR since obviously I'm well acquainted with `nix-build`.)

support/docker/build.sh
```diff
@@ -29,11 +29,136 @@ stderr.printf() {
...
+	fi
+
+
+	# Check if podman is available
+	if ! command -v podman &> /dev/null;
+	then
+		stderr.printf "Podman not installed\n"
+	else
+		availableManagers+=('podman')
+	fi
```
samueldr, line 80: We might want to add to `unavailableManagers` and print the list of unavailable managers when `availableManagers` is empty.

Otherwise we have a spurious non-actionable warning/logging printed.

Or just add to `availableManagers` in an `if` without else and not log it, since the error message currently lists the supported options.

support/docker/build.sh
```diff
@@ -29,11 +29,136 @@ stderr.printf() {
...
+			Quit)
+				stderr.printf "\nSee you soon!\n"
+				exit 0	
+				;;
+			*)
+				stderr.printf "\nInvalid option\n"
+				;;
+		esac
+	done
+
```
samueldr, line 113: If I understand this right, this will *always* stop and ask, right?

This would break the current semantics of acting like a "thick" wrapper to `nix-build` if I got this right.

We could prefer the first in `availableManagers`, if some environment variable is not set.

support/docker/build.sh
```diff
@@ -29,11 +29,136 @@ stderr.printf() {
...
+	
+	# Pull the image if it doesn't exists
+	if (( $checkImg == 0 ));
+	then
+		stderr.printf "\nNixOS image is present\n"
+	elif (( $checkImg == 1 ));
+	then
+		stderr.printf "\nCan\'t find NixOS image. Pulling...\n"
+		$selectedManager pull docker.io/nixos/nix
+	fi
```
samueldr, line 138: Nit: it's the `nix` image, hosted under the NixOS organization. I guess to make this easier to reason about, always printing `nixos/nix` when referring to the image is better, since that is the identifier.

support/docker/build.sh
```diff
@@ -29,11 +29,136 @@ stderr.printf() {
...
+				Yes) 
+					stderr.printf "\nProceeding as root user\n"
+					break
+					;;
+				No)
+					stderr.printf "\nQuiting the build script...\n"
+					exit 0
+					;;
+			esac
+		done
```
samueldr, line 58: Note: This is a case where stopping and asking the user is probably fine? I don't know enough about docker/podman and friends to know if there are drawbacks that warrant stopping running as root.

support/docker/build.sh
```diff
@@ -29,11 +29,136 @@ stderr.printf() {
...
+	declare checkImg
+	case $selectedManager in
+		docker)
+			checkImg="$($selectedManager image ls -a nixos/nix | grep -Fq nixos/nix; echo $?)"
+			;;
+		podman)
+			checkImg=$($selectedManager image exists nix; echo $?)
+			;;
+		*)
+			stderr.printf "\nSomething went wrong when checking if the image is available with the chosen container manager: $selectedManager\n"
```
samueldr:
```suggestion
			stderr.printf "\nUnexpected value for selectedManager ($selectedManager) when checking image is present.\n"
```

This is a message that is unlikely to be seen by a user, but more likely by a developer adding a new manager. Let's make the message precise in what it is about.

support/docker/build.sh
```diff
@@ -29,11 +29,136 @@ stderr.printf() {
...
+	declare volumeExists
+	case $selectedManager in
+		docker)
+			volumeExists=$($selectedManager volume ls | grep -Fq Tow-Boot--nix-store; echo $?)
+			;;
+		podman)
+			volumeExists=$($selectedManager volume exists Tow-Boot--nix-store; echo $?)
+			;;
+		*)
+			stderr.printf "\nSomething went wrong when checking if the volume exists with the chosen container manager: $selectedManager\n"
```
samueldr:
```suggestion
			stderr.printf "\nUnexpected value for selectedManager ($selectedManager) when checking volume exists.\n"
```

This is a message that is unlikely to be seen by a user, but more likely by a developer adding a new manager. Let's make the message precise in what it is about.

samueldr:
Sorry it took this long until I could take a look.

I left some notes about the script.

Though it's possible I missed things. Also note I haven't ran it yet.


One main consideration with the script is it should never\* stop and ask things. Otherwise it'll differ too greatly from the usual `nix-build` commands. It should be a drop-in replacement.

\*: except maybe for that thing about running as root, I don't know.

psstoyanov:
Thank you for the great feedback @samueldr ! 
I will see to block a few hours on Saturday to go through all the notes.

At the top of my head, perhaps an interactive flag could be a solution. It would allow a flexible route when desired by the user while adhering to the "no-stop to ask" rule as a default.

Most guides for starting with Docker and Podman don't add a mention for rootless mode. Most likely to lower the barrier for entry.
Probably a prominent warning that the script is running root mode would be enough.

support/docker/build.sh
psstoyanov: Thank you very much for the suggestion! 
It really makes a huge difference to make the text easy to read.

This is something I should use pretty much everywhere from now on :)

support/docker/build.sh
psstoyanov: Sounds good :+1:

support/docker/build.sh
psstoyanov: Will change it to match the expected image.

support/docker/build.sh
psstoyanov: While it would ideally be a good spot to always request permission, most Docker && Podman starter guides are for root setups.
Will change it to always print something visible. But remove the forced interaction by default.

support/docker/build.sh
psstoyanov: I have to check how to expand this functionality.

I'm thinking of "interactive" mode where the user can do the selections from above.
But to be executed without any required interaction by default.

support/docker/build.sh
ArenM: Does this behave okay in cases where docker is symlinked to podman? As far as I'm aware it's [not uncommon](https://archlinux.org/packages/extra/x86_64/podman-docker/) to have a docker executable that just calls podman.

ArenM (commented):
It looks like progress on this has stalled, would it be helpful if I open up a new pr with a simplified version of this?

Issue #53, opened 2021-09-14
Investigate `upstream_kernel` option for rpi4 users?


colemickens:
https://www.raspberrypi.org/documentation/computers/config_txt.html#upstream_kernel

Not sure if this would help Tow-Boot support both mainline + non-mainline users by having the DTBs for both in places?

But then I'm not sure how Tow-Boot's DTBs would stay in sync with the kernel?

samueldr:
`upstream_kernel` AFAIUI only affects which files gets loaded by the vendor firmware. Having to rely on the vendor firmware's behaviour for this would be a failure with Tow-Boot's goals of making things simpler. And really what this would do is select a specific dtb file that the vendor firmware loads before Tow-Boot is loaded.

Though this does open a parallel question, which is maybe what you really meant:

*Should the dtb filename to load be an "firmware option" configurable by end-users?*

Even one step further:

*Should dtb filename rather be a list of valid names to look for, rather than a static name?* and in that case *Should the preferred dtb filename order be a firmware option configurable by end-users?*

PR #192, opened 2022-09-19
boards/RPi: switch to the u-boot aarch64 combined image

mupuf:
The current version of the raspberryPi-aarch64 image only works on a couple of RaspberryPi boards: 3b and 4b.

Adding support for other boards can be achieved by using u-boot's `rpi_arm64_defconfig`, which supports:

 - Raspberry Pi 3b
 - Raspberry Pi 3b+
 - Raspberry Pi 4b
 - Raspberry Pi 400
 - Raspberry Pi CM 3
 - Raspberry Pi CM 3+
 - Raspberry Pi CM 4
 - Raspberry Pi zero 2 w

This however requires shipping the .dtb files along with the overlays inside the shared image, which this commit also adds.

Unfortunately, the raspberrypifw needs to be updated to achieve this goal, which explains why this commit is marked at WIP:

 - some of the .dtb files are currently missing
 - the start4.elf file is too old for the CM4

### DIFF
--- a/boards/raspberryPi-aarch64/default.nix
+++ b/boards/raspberryPi-aarch64/default.nix
@@ -17,6 +17,12 @@ let
     armstub=armstub8-gic.bin
     disable_overscan=1
 
+    [cm4]
+    dtoverlay=dwc2,dr_mode=host
+
+    [cm4s]
+    dtoverlay=dwc2,dr_mode=host
+
     [all]
     kernel=Tow-Boot.noenv.bin
     arm_64bit=1
--- a/boards/raspberryPi-aarch64/default.nix
+++ b/boards/raspberryPi-aarch64/default.nix
@@ -4,30 +4,21 @@ let
   inherit (config.helpers)
     composeConfig
   ;
-  raspberryPi-3 = composeConfig {
+  raspberryPi-arm64 = composeConfig {
     config = {
-      device.identifier = "raspberryPi-3";
-      Tow-Boot.defconfig = "rpi_3_defconfig";
-    };
-  };
-  raspberryPi-4 = composeConfig {
-    config = {
-      device.identifier = "raspberryPi-4";
-      Tow-Boot.defconfig = "rpi_4_defconfig";
+      device.identifier = "raspberryPi-arm64";
+      Tow-Boot.defconfig = "rpi_arm64_defconfig";
     };
   };
 
   configTxt = pkgs.writeText "config.txt" ''
-    [pi3]
-    kernel=Tow-Boot.noenv.rpi3.bin
-
     [pi4]
-    kernel=Tow-Boot.noenv.rpi4.bin
     enable_gic=1
     armstub=armstub8-gic.bin
     disable_overscan=1
 
     [all]
+    kernel=Tow-Boot.noenv.bin
     arm_64bit=1
     enable_uart=1
     avoid_warnings=1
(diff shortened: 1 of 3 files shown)


Strit:
You should probably rebase against the development branch, instead of the release branch.

samueldr:
(changing the branch brough yet-to-be merged documentation commits, don't worry too much until it's "ready")

Converting to draft since it's WIP.

mupuf:
Thanks @Strit and @samueldr :)

My commit is pretty much ready (aside from the lack of documentation update), but what I don't know how to go about updating NixOS's raspberrypifw package. Is this something I have to ask from NixOS/you guys, or something I can do myself?

samueldr:
Hi, sorry it took a while to come back to this.

> but what I don't know how to go about updating NixOS's raspberrypifw package

If we directly depend on the "thing", we should import the package locally. [Like we're doing with a few things like TF-A and firmwarey bits](https://github.com/Tow-Boot/Tow-Boot/tree/development/support/overlay). I think we should own a copy of the firmware packages to ensure that we are testing specific versions, and that we can upgrade them without causing other changes.

The main things we should rely on from Nixpkgs should be the *toolchain*, e.g. compilers and generic libraries for toolchain use mainly.

mupuf:
> I think we should own a copy of the firmware packages to ensure that we are testing specific versions, and that we can upgrade them without causing other changes.

That sounds like the most sensible approach, indeed! I didn't want to sound rude though, so I did not propose it :p

mupuf:
@samueldr: Sorry for the delay!

I *think* I have done what you wanted me to do, although I must admit that I am wholly unfamiliar with the nix build system. I merely duplicated what you did with the armlogic, and pretty-much copy/pasted the upstream raspberrypi-firmware package (I just kept the relevant files and dropped all the rest).

In any case, this is a nice improvement: it works both on my cm4 and my raspberry Pi 3+

mupuf:
@samueldr: gentle ping :)

oneingan:
Trying to boot a recent nixos iso image: https://hydra.nixos.org/build/211623536/download/1/nixos-minimal-new-kernel-no-zfs-23.05pre460365.3c5319ad3aa-aarch64-linux.iso using your PR in a Rpi B3+.

OS Stage 1 boot hangs in:
```
waiting for device /dev/root to appear...................
Timed out waiting for device /dev/root, trying to mount anyway.
```

EDIT:

More tests:
- same above ISO in a Raspberry Pi 3 Model B with Tow-Boot 2021.10-005 Release works ok,
- same ISO, Rpi 3B with your PR, hangs exactly as in Rpi B3+ (above error).

EDIT 2:

Adding `upstream_kernel=1` to `config.txt` make system boot but weird keyboard behaviour

mupuf:
Thank you for testing the series!

> Adding upstream_kernel=1 to config.txt make system boot but weird keyboard behaviour

`upstream_kernel=1` will be trying to use the upstream dtbs filenames which I believe are missing from `raspberrypi-firmware` ([source](https://www.raspberrypi.com/documentation/computers/config_txt.html#upstream_kernel)). A great read is this issue in the [rpi firmware project](https://github.com/raspberrypi/linux/issues/3237).

Not sure what defaults the tow-boot project should go for (upstream or vendor). I think we should prefer upstream, given that the vendor-recommended way of booting the vendor-recommended kernel would be directly through `config.txt`.

If we go this route, this would mean we need to create another nix package to download kernel-generated dtbs for the RPi, and we'll need to keep them relatively up to date.

Anyway, mind trying to add the upstream dtbs and see if this solves your "weird" keyboard behaviour?

oneingan:
Thanks for reply, some interesting things I've discovered:

I put the mainline dtbs in `upstream/` but they were ignored.

Reading docs seems like `upstream_kernel=1` in Tow-Boot is not loading mainline dts, because `kernel=` is declared. So it only load `upstream.dtbo` from `overlays` directory. I've confirmed this by deleting that file and boot hangs again waiting for `/dev/root`.

Also reading this comment https://github.com/Tow-Boot/Tow-Boot/issues/53#issuecomment-919705431 I'm starting to be confused about how Tow-Boot wants to manage this.

mupuf:
> Thanks for reply, some interesting things I've discovered:
> 
> I put the mainline dtbs in `upstream/` but they were ignored.

Ha, good that you are starting to figure out what is going on!

> 
> Reading docs seems like `upstream_kernel=1` in Tow-Boot is not loading mainline dts, because `kernel=` is declared. So it only load `upstream.dtbo` from `overlays` directory. I've confirmed this by deleting that file and boot hangs again waiting for `/dev/root`.

I see! So, the adaptation overlay is not compatible with newer kernels anymore? I can't say I am surprised, as this definitely feels like a hack.

Would be good if you could figure out if using the upstream dtbs actually solve your keyboard weirdness!

> Also reading this comment [#53 (comment)](https://github.com/Tow-Boot/Tow-Boot/issues/53#issuecomment-919705431) I'm starting to be confused about how Tow-Boot wants to manage this.

I agree with the objective of the project. This BS should not go in the way of users!

According to https://elinux.org/Device_Tree_Linux#forward_and_backward_dts_compatibility, newer kernels need to work with old dtbs, but the opposite is not guaranteed. So, from this PoV, we want the kernel we want to boot to come with its own dtbs (so that they are the most feature-complete), but we *could* also select a kernel that we consider good-enough for most boards and vendor its dtb.

The problem with asking kernels to come with their own dtbs is that, of course, the firmware doesn't know ahead of time which dtbs to load... so that means tow-boot would have to load the right one once it would have selected the kernel to boot. I wonder if there is a way for us to know which board we are running on so that we could be responsible for loading the right dtb (and overlays) at runtime (there must be an RPC for that, but not sure if it is documented).

Additionally, if we can't rely on config.txt to select which overlays to load, we'll have to add a way for users to select which overlays they want to load at boot time, akin to what a BIOS config screen would be. This may be a little too much to ask from tow-boot?

mupuf:
@samueldr: May we ask for your thoughts on this matter?

PR #178, opened 2022-08-21
U-Boot: 2021.10 → 2022.07

samueldr:
Let's finish what I started a small while ago

### TODO

 - [x] Check `XXX` commits
 - [x] Build all devices
 - [x] Test enough representative devices
     - [x] pine64-pinebookA64 (mmcboot installer from blank)
     - [x] pine64-pinebookPro (SPI installer from blank)
     - [x] pine64-rockpro64 [@psstoyanov / @PureTryOut]
     - [x] pine64-pinephoneA64 [@psstoyanov / @PureTryOut]
     - [x] pine64-pinephonePro [@PureTryOut]
     - [x] odroid-C4 [@tpwrules]
     - [x] odroid-N2 [@tpwrules]
     - [x] orangePi-pc
     - [x] pine64-pineA64LTS
     - [x] radxa-zero2
     - [x] libreComputer-rocRk3399PcMezzanine
     - [x] libreComputer-amlS805xAc
     - [x] raspberryPi-aarch64

 

 - *Systems I can't test*
     - [x] radxa-RockPi4 (**rk3399 was spicy this upgrade cycle, might require testing**)
     - [x] radxa-RockPi4C (**same**)
     - [ ] odroid-C2 (Likely fine)
     - [x] pine64-pineA64 (Allwinner is generally boring, should be good)
     - [x] olimex-teresI (Allwinner is generally boring, should be good)
     - [x] orangePi-zeroPlus2H5 (Allwinner is generally boring, should be good)

 - [x] Remove "markers" from commit history

NOTE: systems I can't test are generally covered enough by my representative set of devices. It probably won't be an issue not to test them.

* * *

### Issues noticed

 - [x] ~~upstream regression with rk8xx~~
     - [x] ~~causing regression with USB support with at least the Pinebook Pro~~
 - [x] ~~OS Boot order might be broken again for some scenarios~~
     - [x] ~~Pinebook A64 wouldn't launch installer from installer media startup...~~
     - [x] ~~Pinebook Pro lists SD before eMMC in the boot menu...~~

#### `radxa-zero2`

The mmc/sd device order was previously broken in the initial device enablement in `2021.10-005`. This is a Tow-Boot issue, not an upstream issue, nor a vendor issue.

To upgrade Tow-Boot **from 2021.10** to any further version, first use a 2021.10 `mmcboot` installer, then use the erase function. Once erased, boot the system from SD with the `*-006` or following release of Tow-Boot and install.

[...]

PR #199, opened 2022-10-03
radxa-RockPi4: Fix USB keyboard not working

### LINKED ISSUE #147: rockpi4c keyboard does not work in installer/early boot
I tried a number of different keyboards on all of the usb ports as well as trying through a USB hub. The installer was completely unresponsive.

I (maybe irresponsibly) flashed anyway by modifying the installer image to automatically flash at start. The issue persists when booting Linux (NixOS in my case), where the grub menu cannot be navigated since the keyboard does not work.

I don't have a serial cable, unfortunately, so I can't share the logs from my device.

jirutka:
Note: On ROCK Pi 4B+, the keyboard must be connected to one of the *blue* USB ports.

Fixes #147

### DIFF
--- a/boards/radxa-RockPi4/preboot-usb-start.patch
+++ b/boards/radxa-RockPi4/preboot-usb-start.patch
@@ -1,0 +1,27 @@
+Patch-Source: https://www.mail-archive.com/u-boot@lists.denx.de/msg371891.html
+
+This patch fixes a problem with USB keyboard not working in the boot menu.
+I tested it on ROCK Pi 4 B+ and it works, but the keyboard must be connected
+to one of the *blue* USB ports.
+--
+From: Marcin Juszkiewicz <marcin@juszkiewicz.com.pl>
+Date: Wed, 3 Jan 2020 12:45:26 +0700
+Subject: [PATCH] rockchip: rockpro64: initialize USB in preboot
+
+With video output enabled and USB keyboard supported there is no need
+for serial console. So let initialize USB subsystem so keyboard connected to
+board (either directly or via hub) can be used to interrupt autoboot.
+
+diff --git a/configs/rock-pi-4-rk3399_defconfig b/configs/rock-pi-4-rk3399_defconfig
+index f01b6a3935..b2de4e4cfb 100644
+--- a/configs/rock-pi-4-rk3399_defconfig
++++ b/configs/rock-pi-4-rk3399_defconfig
+@@ -10,6 +10,8 @@ CONFIG_DEBUG_UART_BASE=0xFF1A0000
+ CONFIG_DEBUG_UART_CLOCK=24000000
+ CONFIG_DEBUG_UART=y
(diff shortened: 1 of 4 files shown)


samueldr:
This change AFAICT should only change the *operating system boot order*, it shouldn't affect USB functionality in other ways. Unless for some reason looking for a USB boot device works around a problem.

Have you tested the upcoming update (#178) to see if there is some issues that got solved by upstream? There might be, considering other RK3399 had some USB-related issues fixed by it.

Note that it would be undesirable to prefer booting operating system form USB within Tow-Boot. The order is to first boot internal storage, then external storage, on a "gradient". E.g. eMMC is internal, NVMe is internal, SD is kinda internal, USB is not internal. [though yes, currently rockchip targets would prefer SD over NVMe.]

jirutka:
> This change AFAICT should only change the operating system boot order, it shouldn't affect USB functionality in other ways.

Yeah, it’s weird, but it really fixed the problem with USB keyboard.

> Have you tested the upcoming update (https://github.com/Tow-Boot/Tow-Boot/pull/178) to see if there is some issues that got solved by upstream?

Yes, that was the first thing I did.

> Note that it would be undesirable to prefer booting operating system form USB within Tow-Boot. 

I know, it’s suboptimal but still better than the current state in which Tow-Boot is unusable on this device. I’ll try to find another way though.

jirutka:
Okay, I found a better way, `CONFIG_PREBOOT="usb start"`. I’ve also backported a patch from u-boot (https://github.com/u-boot/u-boot/commit/d5daa02d8d9e7c403a3339db1966e8413e64e408) to fix a reboot loop occuring with some USB devices.

samueldr:
Notes to self: the changes to `drivers/usb/host/xhci-ring.c` are part of 2022.07, so the patch is not needed.

The `radxa-rockpi4` still doesn't enable PREBOOT with `usb start`, so this will be needed to be addressed.

boards/radxa-RockPi4/preboot-usb-start.patch
```diff
@@ -0,0 +1,27 @@
+Patch-Source: https://www.mail-archive.com/u-boot@lists.denx.de/msg371891.html
+
+This patch fixes a problem with USB keyboard not working in the boot menu.
+I tested it on ROCK Pi 4 B+ and it works, but the keyboard must be connected
+to one of the *blue* USB ports.
+--
+From: Marcin Juszkiewicz <marcin@juszkiewicz.com.pl>
+Date: Wed, 3 Jan 2020 12:45:26 +0700
+Subject: [PATCH] rockchip: rockpro64: initialize USB in preboot
```
samueldr, line 9: sneaky!

I found "this patch", and got confused seeing it was accepted and merged in 2020.

Note to anyone missing the sneaky bits: this is an edited patch for another board.

We probably should instead move this to Tow-Boot config, since as-this-is it won't be inherited by the 4C, since it uses a different defconfig.

(Still notes to me)

PR #213, opened 2022-11-13
libreComputer-amlS905xCc: Add support for AML-S905X-CC (Le Potato)

samueldr:
This tentatively adds a build for the AML-S905X-CC. I don't have the hardware, but have someone handy who is curious and interested.

cc @josqu4red

cc https://github.com/NixOS/nixpkgs/pull/200863

### DIFF
--- a/boards/libreComputer-amlS905xCc/default.nix
+++ b/boards/libreComputer-amlS905xCc/default.nix
@@ -1,0 +1,21 @@
+{ config, lib, pkgs, ... }:
+
+{
+  device = {
+    manufacturer = "Libre Computer";
+    name = "Le Potato";
+    identifier = "libreComputer-amlS805xAc";
+    productPageURL = "https://libre.computer/products/aml-s905x-cc/";
+  };
+
+  hardware = {
+    soc = "amlogic-s905x";
+  };
+
+  Tow-Boot = {
+    defconfig = "libretech-cc_defconfig";
+    builder.additionalArguments = {
+      FIPDIR = "${pkgs.Tow-Boot.amlogicFirmware}/lepotato";
+    };
+  };
+}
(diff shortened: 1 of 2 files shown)


josqu4red:
Hi @samueldr, thanks for this!

I just tested it only with `shared.disk-image.img` written to the SD card, seems to be working properly.
I've yet to succeed with building an actual OS image to really validate my nixpkgs/uboot change and this one though.

samueldr:
Try booting the stock UEFI NixOS image from USB from Tow-Boot [or stock U-Boot].

Amlogic "this old" (not that old) can generally boot just fine on mainline Linux.

You can look here for some docs and links for the iso:

 - https://nixos.wiki/wiki/NixOS_on_ARM/UEFI#Getting_the_installer_image_.28ISO.29

Though do note that care must be taken in installing to *not* format with GPT and keep the existing MBR, because of *Amlogic idiosyncrasies*.

(Also any other recent mainline Linux UEFI iso should work.)

josqu4red:
Tried with Nixos aarch64 on USB without success, neither this tow-boot build nor the u-boot one from the other PR.
System crashed with `"Synchronous Abort" handler` seemingly when Grub tries to boot the kernel.
From there, I guess the issue is more with the image than bootloader, but I have no idea how to verify that.

Boot log attached in case you want to take a look
[towboot.log](https://github.com/Tow-Boot/Tow-Boot/files/10170618/towboot.log)

samueldr:
> seemingly when Grub tries to boot the kernel

Thank you for the log, it is the right info to guesstimate.

Did you have the GRUB menu show on screen? Did you have HDMI out connected?

> I guess the issue is more with the image than bootloader, but I have no idea how to verify that.

Try fedora, or another known good UEFI-friendly AArch64 distro.

josqu4red:
Fedora is booting fine with either distributions!
Though now I have to figure out how to make nixos work here :grimacing:

Arvind-Jana (commented):
Good

PR #220, opened 2022-12-26
u-boot 2022.10 created, forking off of u-boot 2022.07

IreneKnapp:
this has some duplication - if you'd like me to reorganize the patches to avoid that, just let me know how you'd like them. the only patch that isn't identical to its 2022.07 version is 0001-pdcurses.


Strit:
Since this was made even u-boot 2023.01 was released.

samueldr:
This will be closed with #238 changing how U-Boot updates are handled, but since I intend to go up a one release at a time to more easily deal with regressions and other issues, I'll try and use any of the equivalent patches that needed changes with co-author credit.

PR #221, opened 2022-12-29
Fix link to ROCK Pi 4 board

axelsimon:
Hope this helps and i haven't completely misunderstood how the devices's READMEs are designed to refer to one another :)

### DIFF
--- a/boards/radxa-RockPi4C/README.md
+++ b/boards/radxa-RockPi4C/README.md
@@ -2,5 +2,5 @@
 
 ## Device-specific notes
 
-This board is nearly identical to the [Radxa ROCK Pi 4 A/B](./devices/radxa-RockPi4.html).
+This board is nearly identical to the [Radxa ROCK Pi 4 A/B](../radxa-RockPi4).
 As a result, all device-specific notes from there also apply here.


samueldr:
Hi! Sorry it took a while until I took a look here.

So this is a case of mismatched expectations, and will have to be handled (you don't need to, I understand bespoke tooling is a drag to work on) in a different way.

So the underlying problem is that this is a case of mismatched expectations between the built documentation website, and the github linking between pages.

If you use the website, the current link works:

 - https://tow-boot.org/devices/radxa-RockPi4C.html

And yes, this is entirely incorrect for navigation within GitHub.

The solution here will be to look into rewriting the "github-correct" links when building the website, when cross-referencing between board is added... Though that alone won't be enough as other cross-links within the docs won't be correct...

More as a note to myself: this is already handled for non-board links for `.md` → `.html` [here](https://github.com/Tow-Boot/Tow-Boot/blob/00b18a0a85cc047425ce0f4a512b7ff3bb0571dc/doc/_support/converter/main.rb#L139).

samueldr:
(Closing, since the proper fix will be entirely different, and an issue is open for tracking that now.)

samueldr:
The fix ended-up being cherry-picked in the project; I made a mistake analyzing the github link requirements, and couldn't use what I had in mind. Though it couldn't have gone in *alone*, as it would have broken the link in the generated documentation site.

See https://github.com/Tow-Boot/Tow-Boot/pull/243

PR #226, opened 2023-02-03
Add imx8qm support.

unbel13ver:


samueldr:
Hi! :wave: 

First of all, sorry for taking a while until coming back to this PR.

At first glance a lot of it looks okay, like your comments about needing to generalize: when adding a new family, it often happens that we don't have enough knowledge floating about to correctly generalize, so as long as the inputs and outputs are correct enough to handle generalizing later without much pain, I think it's likely fine.

Since this is adding a new family outright, I'm pushing this in the next release.

I will also evaluate and compare with the WIP-yet-untested branch I had for the Librem 5, which built with what looked like *probably* working outputs. And then try to guesstimate some of the generalization and get back around to you for seeing if you can judge whether the changes are okay or not.

Though one really problematic part here: reliance on a vendor fork of U-Boot. That will require thinking about how to handle: we don't have the capacity to track more than one upstream. *Ideally*, we'd see how that is custom is ***needed*** for a support level similar to the other boards (e.g., no fancy TF-A, OP-TEE) and see if forward porting to mainline and maybe working with them to contribute that subset would be feasible. But that's something to evaluate later on.

PR #228, opened 2023-03-12
Refactor raspberryPi board configuration

oneingan:
I mix and match previous community attempts to update `raspberrypi-aarch64` board stack. To summarize:

- I have created a hardware freeform module to configure global options and filters in `config.txt.`
- Make module options for important options (by the moment `upstreamKernel`).
- It creates the fw filesystem with both `rpi` and `mainline` kernel files.
- Use new unified U-Boot kernel for all `arm64` raspberries.

I, obviously, have been inspired by @colemickens and @mupuf, as in:
https://github.com/Tow-Boot/Tow-Boot/pull/192
https://github.com/colemickens/Tow-Boot/tree/rpi/boards/raspberryPi-aarch64

Also, the @colemickens setup was the only one that works flawlessly for me using mainline kernel in both RPi 3 and RPi 3B+. But was hard to use their modifications directly. That was the main reason I had to try to apply their improvements upstream.

Eventually superseed also #156 and #61


mupuf:
Oh oh, that looks very interesting!

I like the fact that you sidestepped the issue of not knowing if the kernel was going to be an upstream one or not by just... putting it on the user to select via an option menu! I mean, this is not fully what we aspire for here, but this is a nice workaround until we do achieve (which would require reimplementing part of what the firmware is doing)! No need to rebuild/reflash the bootloader because of needlessly-opiniated defaults :)

I will give it a try during the week, and close my pull request. Thanks a lot!

modules/hardware/raspberryPi/default.nix
```diff
@@ -0,0 +1,270 @@
...
+              ## rpi firmware DTBs
+
+              ## rpi4 armstubs
+              cp -vt "$target/" "${cfg_rpi.armstubsPackage}/armstub8-gic.bin"
+
+              ## mainline (kernel, dtbs, !overlays)
+              mkdir -p "$target/upstream"
+              cp -vt "$target/upstream/" \
+                "${config.Tow-Boot.outputs.firmware}/binaries/Tow-Boot.noenv.bin" \
+                ${cfg_rpi.mainlineKernel}/dtbs/broadcom/bcm*rpi*.dtb
```
mupuf, line 234: Do we really need to compile the whole kernel to generate the dtb? I mean, we definitely could pick a kernel, download all the device trees and compile them ourselves.

mupuf:
Can you update the firmware version too? My CM4 doesn't want to boot with the image I generated using your MR.

Feel free to use my code for it!

modules/hardware/raspberryPi/default.nix
oneingan: Sure, I didn't have this problem because my build hit cache (native compile). Just check that you are right and crosscompiling implies build whole kernel.

oneingan:
> Can you update the firmware version too? My CM4 doesn't want to boot with the image I generated using your MR.
> 
> Feel free to use my code for it!

Checking your code, I am wondering if use an additional updated nixpkgs ref (only for this) could be best for ease of maintenance.

samueldr:
(This milestone is not a judgement of the PR; I will take a look later. It's only to sort things out in their proper bucket.)

PR #238, opened 2023-06-26
Drop forest of patches, and use the new unified tree

samueldr:
This change set aims to make it easier to work and collaborate on the patches.

This makes use of the (Tow-Boot/U-Boot)[https://github.com/Tow-Boot/U-Boot] repo, where topic branches are created for releases.

 - https://github.com/Tow-Boot/U-Boot/branches/all?query=tow-boot%2F2022.07

A final strategy hasn't been decided for dealing with the actual branches, but the gist of it is once "ready", a "final" unified branch is pushed, which is an octopus merge of the whole set of changes. This merge is made using a tool, and *should* also be used when validating work before publishing things for U-Boot changes. Note that this is still a WIP and likely will change, TBD with other collaborators.

Though what is final is that we will be using a single source to represent the whole change set<b>\*</b>.

One benefit that has already been shown, where some *way too hacky* changes were in use, and it instead was preferable to work things out in a better manner with discrete U-Boot config options. In other words, it keeps us more honest and in better working order, without giving us as much opportunities to lapse into bad hacks.

The code, short of some minor cleanups, mainly centered around surrounding hacks with `CONFIG_TOW_BOOT*` options, the code is basically identical. I have verified that by taking a snapshot (new git repo) of the source tarball produced by the current `development` tip, then this PR, for the same board (`pine64-pinebookA64`), and looking at the [diff](https://gist.github.com/samueldr/99053e0964e358256e7f36f3e54098df). This is "hard" to judge properly because now all the patches are included all the time, but the changes, other than new boards and new features (unused) are minor and trivial.

### What was done

 - [x] Build a full release (all boards)
 - [x] Fixup the build so building stock U-Boot with the tooling works
 - [ ] Actually run on a representative subset of hardware
     - [x] Tested on pine64-pinebookA64 (representative of Allwinner)
     - [x] Fix PBP panel reset quirk patch <del>I</del> that broke
         - [x] Re-work RK3399 wip patches out of Pinephone branch
     - [x] Re-test on PBP (SPI flash already tested successful, works for everything <del>except panel on reboot</del>)

### What needs to be done

 - [ ] Actually run on a representative subset of hardware
     - [ ] Amlogic
     - [ ] Raspberry Pi
     - [ ] Pinephone Pro (testing both revised RK3399 changes, and the Phone UX)

### Out of scope of this PR

 - Figuring out the collaboration strategy. Whichever it is, it will desugar down into a `fetchFromGitHub` to a repo, shared between all boards.

* * *

> \* This is almost true, we still do some patching down the line for some things, since it needs to happen for a U-Boot build made using our tooling too. Namely, adjusting SPL offset for SPI builds, and serial speed on rockchip. This should be assumed work to somehow prepare in a somewhat upstreamable fashion.


modules/tow-boot/config.nix
```diff
@@ -119,6 +119,10 @@ in
       TPL_ENV_IS_NOWHERE = option yes;
       SPL_ENV_IS_NOWHERE = option yes;
     }))
+    (mkIf (variant == "boot-installer") (helpers: with helpers; {
+      TOW_BOOT_PREDICTABLE_BOOT_PREFER_EXTERNAL = yes;
+      TOW_BOOT_PREDICTABLE_BOOT_PREFER_INTERNAL = no;
```
samueldr, line 125: Since this is a `choice`, just the `PREFER_EXTERNAL = yes;` should be enough, but having a `no` for the other *ensures* there's nothing wrong going on with the implementation. It's redundant, but gives assurance both won't be true at once.

modules/tow-boot/config.nix
```diff
@@ -69,7 +69,7 @@ in
       CMD_BDI = yes;
       CMD_CLS = yes;
       CMD_SETEXPR = yes;
-      CMD_PAUSE = yes;
+      CMD_PAUSE = lib.mkIf (!config.Tow-Boot.buildUBoot) yes;
```
samueldr, line 73: `CMD_PAUSE` has been mainlined, so it will be possible to enable this outright when upgrading U-Boot.

samueldr:
For anyone playing at home:

 - Making the unified tree broke the display reset workaround
 - The code is unchanged
 - The DT for rk3399 was synced
     - https://github.com/Tow-Boot/U-Boot/commit/63324713813b823d0d6d8ae9d660efe42952bc44
 - The DT node names changed
     - https://github.com/Tow-Boot/U-Boot/commit/63324713813b823d0d6d8ae9d660efe42952bc44#diff-c3bedab72d65b361172dbc07654eb3c8ac9fbda871fdc2ef9eb2f058af141fbfL1961
 - The workaround uses `gpio_lookup_name`
     - https://github.com/Tow-Boot/U-Boot/commit/cba7cea410780bfcdd72b5d8ae9c190c0b1572ba#diff-1d2e4da763e7572659b72185b8c8216ac91e572573d4915224c421bbce259e67R76
 - This, in turn, relies on the *node name*, which now is not differentiated.

This shows a *correctness* issue when syncing DTs with linux and that workaround. From a quick grep, no other `gpio_lookup_name` exist for rk3399. So we're safe keeping the sync in.

The solution is to change the workaround implementation.

This turns a regulator off and on again. This is not elegant. This also shows garbage in the display for a short while.

What we can do, instead, is force the regulator off during SPL. Since it's `always-on`, it will be turned on at boot. Note that it will be turned on *just like from a cold start*. So we can prod it the same way LEDs are prodded in SPL.

 - https://github.com/samueldr/u-boot/commit/fc5bee948ae3b8bfc3ca36a2dd40e620cffcc845

* * *

A better(?) solution might be to have a "for U-Boot" tag on the regulator telling it that actually while it is always-on, SPL should turn it off, as a warm reset keeps the thing in a bad state. Or maybe it shouldn't be `always-on`?

With the upcoming change, there is no *garbage* on display, but the framebuffer <del>was not cleared, so it shows the last render from the platform firmware (when resetting from the platform firmware). This would be an issue with the `rk_edp` driver not clearing the memory, and/or maybe the backlight should be turned on later so the display is actually reset first(?)</del> is now cleared.

<del>(Fix WIP, will be pushed to the unified tree soon, as it requires revamping the organization of the Pinephone Pro changes, RK3399 changes, and some minor cleanups in the Pinebook Pro changes.)</del>

samueldr:
Latest push updates the source ref; which in turn has the RK3399 history revisions, Pinebook Pro revisions, and framebuffer clearing.

 - https://github.com/Tow-Boot/U-Boot/compare/f9a5907338ddd9cd46a51781df5300b40c9a8127..66559169abb4558d5d4c60d98a719fdbce315614

samueldr:
(Force-push synced with current `development` branch)

samueldr:
*CI is already paying dividends*

 - https://github.com/Tow-Boot/Tow-Boot/actions/runs/5491031789/jobs/10007233414

samueldr:
It wasn't entirely "tested" in full, like I wanted initially, but the reasoning to merge is two-fold:

 - I am already building up on top of this for the U-Boot update
 - Any platform-specific fixes requires work outside this repo anyway

Nothing here will require "fixing" *other than the revision and hash* once that is done.

Which leads to one question: how should that be managed for updates? I guess we'll add a pin in only when ready to publish a new version?

PR #246, opened 2023-07-06
Make it work in pure evaluation mode

aciceri:
Otherwise Tow-Boot cannot be used from flakes.


samueldr:
Just noting that Tow-Boot *shouldn't* be used from flakes, or even any other automated build system. It's supposed to live outside of the lifecycle of an operating system; installed and managed independently.

Though, with that said, this is a correctness fix. Pure evaluation should work, and this is obviously a mistake. Why pick `pkgs` from the pinned input in the repo there, when it's also a default value for it at the top? Obviously an oopsie.

Thank you!

PR #250, opened 2023-07-11
arm-trusted-firmware: 2.6 -> 2.9

ArenM:
I'm not sure why exactly, but this seems to be needed to get cpuidle working on pine64-pinephoneA64.

I've tested on the pinephone to make sure it still boots from emmc and sd and that cpuidle is now working (with the hacks that distros are using), but I don't have the hardware to test anything else.

I also removed sptool from the armTrustedFirmwareTools package because it's been rewritten in python, requiring adding extra dependencies, and it doesn't appear to be used.

### DIFF
--- a/support/overlay/arm-trusted-firmware/default.nix
+++ b/support/overlay/arm-trusted-firmware/default.nix
@@ -15,7 +15,7 @@ let
             , platform ? null
             , extraMakeFlags ? []
             , extraMeta ? {}
-            , version ? "2.6"
+            , version ? "2.9"
             , ... } @ args:
            stdenv.mkDerivation ({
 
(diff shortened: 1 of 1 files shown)


samueldr:
Aww, I hate being the bearer of bad news, but I have those changes already cooking among other bits:

 - https://github.com/samueldr/Tow-Boot/commits/wip/updating-u-boot

There is also an update for `crust`, and Nixpkgs too. I suspect the crust update could also help with suspend and similar stuff.

ArenM:
Oh I missed that, I'll close this since you seem to have all the same changes there.

I'd hardly call it bad news that you're already working on it :)

samueldr:
No worries, you *couldn't have known*, since it's not a PR yet :)

PR #252, opened 2023-07-16
Add stock U-Boot refs up to 2023.07

samueldr:
**This does not update Tow-Boot.**

* * *

This adds the missing U-Boot versions up to 2023.07 to the knowledge for stock U-Boot.

This also fixes some missing version-gating of configs.

And finally handles some required patching.

* * *

### What was done

 - Built U-Boot sandbox with `nix-build support/check-all-stock-uboot.nix`
 - Built Pinebook A64 for 2022.07..2023.07
 - Ran 2023.07 on the Pinebook A64

For our purpose here, this is sufficient.

Reminder that this is supposed to serve as a way to verify regressions with stock U-Boot with a limited scope. And generally helps us confirm when things break if it's Tow-Boot or U-Boot. It is a development tool and not a product for end-users.

### DIFF
--- a/modules/tow-boot/builder.nix
+++ b/modules/tow-boot/builder.nix
@@ -180,9 +180,11 @@ in
           ] ++ makeFlags;
 
           # Inject defines for things lacking actual configuration options.
-          NIX_CFLAGS_COMPILE = optionals withLogo [
-            "-DCONFIG_SYS_VIDEO_LOGO_MAX_SIZE=${toString (1920*1080*4)}"
-          ];
+          NIX_CFLAGS_COMPILE =
+            (optionals withLogo (
+              lib.optional (lib.versionOlder uBootVersion "2023.01") "-DCONFIG_SYS_VIDEO_LOGO_MAX_SIZE=${config.Tow-Boot.VIDEO_LOGO_MAX_SIZE}"
+            ))
+          ;
 
           extraConfig = ''
             #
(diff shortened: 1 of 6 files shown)


modules/tow-boot/builder.nix
```diff
@@ -180,9 +180,11 @@ in
...
 
           # Inject defines for things lacking actual configuration options.
-          NIX_CFLAGS_COMPILE = optionals withLogo [
-            "-DCONFIG_SYS_VIDEO_LOGO_MAX_SIZE=${toString (1920*1080*4)}"
-          ];
+          NIX_CFLAGS_COMPILE =
+            (optionals withLogo (
+              lib.optional (lib.versionOlder uBootVersion "2023.01") "-DCONFIG_SYS_VIDEO_LOGO_MAX_SIZE=${toString (1920*1080*4)}"
+            ))
+          ;
```
samueldr: - [x] Add a discrete (internal) config in Tow-Boot, use it here, and configure the new Kconfig option.

rik-shaw:
Completely ignorant question, but I have a Pinebook Pro with Tow-Boot. To get sleep (s2idle I think) to power down correctly, I understand I need an updated u-boot. Can I do this with Tow-Boot or do I need to remove Tow-Boot? I could post this as an "issue" if better. Honestly I am confused :-)

samueldr:
Please post a new issue, and add sources explaining what you're talking about with sleep.

Also explain the issues you're observing in details.

Note that a Tow-Boot update to 2023.07-xxx will be coming up in the not-distant future, it's mostly working in a WIP branch, testing needs to be done, and solving a medium-hard problem for rockchip systems is too.

Issue #209, opened 2022-11-05
2022.07-development: USB problems on libreComputer-rocRk3399Pc


samueldr:
> > I have added fixes that should now fix USB with the 2022.07 upgrade.
> 
> Just tested "feature/u-boot-2022.07" and "development" branches. On both branches USB Type-A ports works now (I didn't tested Type-C ports yet) although both throw USB-related errors in both cases, even if no USB devices is attached ("Failed to get clocks"). However, there is severe regression on development branch (I flashed it to SPI) - right after "Please press ... to enter the boot menu." Tow-Boot starting to scan mmc, print stack trace and then reboot. I can get into the menu if I press Esc soon enough, but even if I choice USB I still getting same outcome. Yet somehow spiinstaller works fine, I seeing issue only when I have Tow-Boot installed in SPI (shared image was not tested).

_Originally posted by @shvetsnikita in https://github.com/Tow-Boot/Tow-Boot/issues/178#issuecomment-1304636360_

samueldr:
A few things to check, which config are you using?

 - libreComputer-rocRk3399Pc
 - libreComputer-rocRk3399PcMezzanine

I should `diff` the configs, upstream hasn't been keeping them in sync.

Have you tried erasing SPI flash entirely? I'm asking since if you e.g. had saved the environment to SPI flash, *maybe* that could have saved values use within some drivers or elsewhere that are now wrong... Environment use in U-Boot isn't really great.

Are the crashes happening only with USB devices connected? Which? Tried different combinations?

samueldr:
(Note to self, read https://github.com/Tow-Boot/Tow-Boot/issues/183 more closely and see if it is actually related)

shvetsnikita:
> which config are you using?

Without Mezzanine.

> Have you tried erasing SPI flash entirely?

Just tried to erase SPI entirely, then flash same build again - no any changes in behavior (still USB errors even when no USB devices is attached, even more error if USB dock is attached, crash and reboot after "Please press" message.

> Are the crashes happening only with USB devices connected? 

No, it crash without USB devices too.

samueldr:
Thanks for answering. These mainly help me determine what I need to test.

I am testing with the mezzanine config, since *in theory* they should have been the same, except for the dtb name built into the config. Right now, as-it-is, there seems to be no differences that should matter in the configs. The main difference in the config seems to be related to PCI, all of them.

The dts *seem* to only differ with features relevant to the mezzanine board.

[...]

PR #255, opened 2023-07-22
WIP: Update base to U-Boot 2023.07

samueldr:
Feature branches here:

 - https://github.com/Tow-Boot/U-Boot/branches/all?query=tow-boot%2F2023.07%2F

### TODO

We'll need to tag the current state of the Tow-Boot fork, whether or not it is what is released as `2023.07`.

 - [x] Tag `2023.07-007-rc1` in the U-Boot repo with the accepted state
 - [x] Change impl. to work with that tag while in rc

### Known issues

<details><summary>"Resolved" rockchip issue</summary>

#### Rockchip

Rockchip moved to "standard[sic] boot" breaking the assumptions of the boot menu builder.

AFAIK this is the only actual real-world target/family that has moved to it.

A solution will need to be found, I haven't spent much time thinking about it yet.

The first thought I had is that *since anyway* we have to mess with distroboot things for most targets, we might benefit from making a "unified" distroboot env within our fork, instead of relying on the board-specific distroboot builders. And on top of that, making the *order* they are tried a `CONFIG_TOW_BOOT_DISTROBOOT_TARGETS` option, meaning that the boot order and targets can be handled through direct configuration rather than messy patches and `#ifdef`s.

That solution would be a stop-gap until the new boot methods are more universally enabled (maybe we can help down the line) and we would instead use the data from bootstd directly without going through cli commands when building the UI.

</details>

* * *

### Changelog

 - "Resolved" the Rockchip conundrum https://github.com/Tow-Boot/U-Boot/compare/d705eb1a60dbc6d507974afbf1c26663c321077c...61f7ac4831c93f74f2f542966d072f4aa62fc5cc

### DIFF
--- a/boards/uBoot-sandbox/default.nix
+++ b/boards/uBoot-sandbox/default.nix
@@ -24,6 +24,9 @@
         VIDEO_SANDBOX_SDL = yes;
         SANDBOX_RAM_SIZE_MB = freeform "512";
       })
+      (helpers: with helpers; {
+        BOOTSTD = lib.mkForce yes;
+      })
     ];
     builder = {
       buildInputs = [
--- a/embedded-linux-os/devices/pine64-pinephoneA64/kernel/default.nix
+++ b/embedded-linux-os/devices/pine64-pinephoneA64/kernel/default.nix
@@ -6,7 +6,7 @@ stdenv.mkDerivation rec {
   version = "5.15.0";
 
   src = fetchFromGitHub {
-    owner = "megous";
+    owner = "samueldr"; # Originally megous
     repo = "linux";
     rev = "3cc817fa5bfb1f9c6c1630707cd7aa6d00f4efe3";
     sha256 = "0hwi3z7sppw9pnxjjy0skrrgjicv652k6mlzz1q3nkimbfdgm6cs";
--- a/modules/hardware/rockchip/default.nix
+++ b/modules/hardware/rockchip/default.nix
@@ -16,6 +16,7 @@ let
     variant
   ;
   cfg = config.hardware.socs;
+  withSPI = config.hardware.SPISize != null;
 
   firmwareMaxSize = 4 * 1024 * 1024; # MiB in bytes
   partitionOffset = 64; # in sectors
@@ -46,7 +47,7 @@ in
     (mkIf cfg.rockchip-rk3399.enable {
       system.system = "aarch64-linux";
       Tow-Boot = {
-        config = [
+        config = mkIf withSPI [
           (helpers: with helpers; {
             # SPI boot Support
             MTD = yes;
--- a/modules/tow-boot/builder.nix
+++ b/modules/tow-boot/builder.nix
@@ -122,6 +122,11 @@ in
             patchShebangs tools
             patchShebangs arch/arm/mach-rockchip
           '' +
+          ''
+            # Drop that from the exposed version, always.
+            # We use releases, any extra qualifier is owned by us.
+            sed -i -e 's/^EXTRAVERSION =.*/EXTRAVERSION =/' Makefile
+          '' +
             # FIXME: review how we patch this out... (I don't like it)
           ''
             echo ':: Patching baud rate'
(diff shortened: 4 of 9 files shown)


shvetsnikita:
From my quick testing on a couple of already supported boards (Renegade Elite and OrangePi PC) I noticed that when Tow-Boot ask to press Esc or Ctrl+C, keyboard doesn't work at all (even no reaction to NumLock, so not powered?) but in GRUB2 keyboard works fine.

shvetsnikita:
Another observation is that SPI install doesn't work on Renegade Elite.

shvetsnikita:
At least on Renegade Elite Linux does not detect USB controllers (lsusb is empty) unless I use kernel's own dtb. Relevant messages in dmesg:

```
[    5.216389] usbcore: registered new interface driver usbfs
[    5.216510] usbcore: registered new interface driver hub
[    5.216625] usbcore: registered new device driver usb
[    5.689858] OF: graph: no port node found in /i2c@ff160000/usb-typec@22
[    5.897673] OF: graph: no port node found in /i2c@ff3d0000/usb-typec@22
[   16.106883] platform fe800000.usb: deferred probe pending
[   16.106894] platform fe900000.usb: deferred probe pending
[   16.106905] platform fe3a0000.usb: deferred probe pending
[   16.106913] platform fe3e0000.usb: deferred probe pending
[   16.106922] platform fe380000.usb: deferred probe pending
[   16.106930] platform fe3c0000.usb: deferred probe pending
[   16.106939] platform ff770000.syscon:usb2phy@e450: deferred probe pending
[   16.106947] platform ff770000.syscon:usb2phy@e460: deferred probe pending
```

samueldr:
Can you check for your issues against

 - the current stable release (2022.07-006)
 - stock U-Boot [2022.07, 2023.07]

Reporting a matrix of where things break and fail under as much of the same conditions would help figure out if it's a regression, and if it is, where it may come from.

shvetsnikita:
> Reporting a matrix of where things break and fail under as much of the same conditions would help figure out if it's a regression, and if it is, where it may come from.

Tow-Boot on Renegade Elite:
- SPI installer works on Tow-Boot 2022.07, but not on Tow-Boot 2023.07 - installer fail to perform writing image to SPI. 
- If I press Esc (on directly attached USB keyboard) when Tow-Boot 2022.07 propose to do so, it'll drop me to the menu. If I press Esc when Tow-Boot 2023.07 propose to do so, I either getting no response (Tow-Boot proceed with boot of installed OS from eMMC) or it'll drop me to the console, but never to the menu.
- If I boot Linux 6.1 (Debian stable) or 6.4 (Debian unstable) with Tow-Boot 2022.07 - USB works (albeit with [some issues](https://github.com/Tow-Boot/Tow-Boot/issues/209#issuecomment-1660388599), but overall it works) but if I boot same kernels with Tow-Boot 2023.07 - I getting [USB-related errors in dmesg](https://github.com/Tow-Boot/Tow-Boot/pull/255#issuecomment-1660391979) and empty lsusb. It's possible to workaround this specific issue with Tow-Boot 2023.07 by supplying kernel's own dtb from /lib/firmware via grub "devicetree" command - then USB works on Tow-Boot 2023.07 too.

Upstream U-Boot on Renegade Elite (I used [this project](https://github.com/johang/sd-card-images) to build img):
- With U-Boot 2022.07 USB doesn't work in U-Boot and GRUB2 (no even directly attached keyboard) but works in Linux.
- With U-Boot 2023.07 USB works in U-Boot and GRUB2, but not in Linux, just like with Tow-Boot 2023.07

So not working USB in Linux is seems like upstream regression. As for SPI installer and menu in Tow-Boot 2023.07 - I guess it's Tow-Boot own regressions?

samueldr:
Oh, the menu not working *may* be related to the move off of distroboot for rockchip hardware; the Tow-Boot menu is built using distroboot semantics, and this is the main thing we need to fix. So for any rk3399 targets, I'd say wait until that has been "finished" on my end.

And yes, if mainline U-Boot shows the same regression in equivalent builds, your assumption here is right: it's likely something caused within U-Boot.

PR #256, opened 2023-07-31
Add support for Libre Computer ALL-H3-CC H5

shvetsnikita:
Support was tested only with [this tree](https://github.com/samueldr/Tow-Boot/tree/feature/tow-boot-2023.07).
At least Libre Computer [own image](https://distro.libre.computer/ci/ubuntu/23.04/ubuntu-23.04-preinstalled-desktop-arm64%2Ball-h3-cc-h5.img.xz) boot fine.

### DIFF
--- a/boards/libreComputer-allH3CcH5/default.nix
+++ b/boards/libreComputer-allH3CcH5/default.nix
@@ -1,0 +1,16 @@
+{
+  device = {
+    manufacturer = "Libre Computer";
+    name = "ALL-H3-CC H5";
+    identifier = "allH3CcH5";
+    productPageURL = "https://libre.computer/products/all-h3-cc/";
+  };
+
+  hardware = {
+    soc = "allwinner-h5";
+  };
+
+  Tow-Boot = {
+    defconfig = "libretech_all_h3_cc_h5_defconfig";
+  };
+}


boards/libreComputer-allH3CcH5/default.nix
```diff
@@ -0,0 +1,15 @@
...
+    productPageURL = "https://libre.computer/products/all-h3-cc/";
+  };
+
+  hardware = {
+    soc = "allwinner-h5";
+  };
+
+  Tow-Boot = {
+    defconfig = "libretech_all_h3_cc_h5_defconfig";
+  };
```
samueldr, line 15:
```suggestion
  }
```

shvetsnikita:
Just noticed that board is [supported by crust](https://github.com/crust-firmware/crust/blob/master/configs/libretech_all_h3_cc_h5_defconfig), so maybe it's good idea to include it?
```
  hardware = {
    soc = "allwinner-h5";
    allwinner.crust.defconfig = "libretech_all_h3_cc_h5_defconfig";
  };
```

samueldr:
Definitely yes, if it is verified to work (or at least not break if you're unsure how to confirm it's working 100%).

shvetsnikita:
> or at least not break if you're unsure how to confirm it's working 100%

I didn't noticed any difference between Tow-Boot build based on previously submitted PR and Tow-Boot build with enabled crust support. This time I tested not only Libre Computer own image, but also regular Debian 12 installed on flash drive, each one was tested with crust-enabled Tow-Boot and no-crust Tow-Boot build.

I also noticed that this board support eMMC, but I don't have compatible eMMC module to test Tow-Boot eMMC install.

samueldr:
Force-pushed, updating to the current state of `development` (which brings in the CI), and with the two fixups squashed.

Also dropped (in the next force push) the tautological enablement of crust. Crust is automatically enabled and assumed to use the `defconfig` name that U-Boot does, because for most devices it is all that is needed.

PR #257, opened 2023-08-01
Add support for FriendlyELEC NanoPi M1

shvetsnikita:
Support was tested only with [this tree](https://github.com/samueldr/Tow-Boot/tree/feature/tow-boot-2023.07).

At least Debian 12 armhf boot fine, but in my case - only in headless mode, which could be related to my display (Dell UP3017) resolution 2560x1600. However, while I getting black screen when kernel switch from framebuffer, video output before this moment (including video output in Tow-Boot and GRUB2) works without issues. Another observation is that USB support is a bit flaky - Tow-Boot goes into boot loop if Belkin USB-C 3.1 Express Dock HD or even Satechi Type-C Card Reader (expose two block devices) is connected, but this isn't likely to cause any issues in real-world scenarios for this board (signage, thin client, print server, pihole, etc.)

Anyway, with 2022.07 test branches situation was even worse (I wasn't able to even get to Tow-Boot's menu as I remember) so we still have a nice progress here.

### DIFF
--- a/boards/friendlyElec-nanoPiM1/default.nix
+++ b/boards/friendlyElec-nanoPiM1/default.nix
@@ -1,0 +1,17 @@
+{
+  device = {
+    manufacturer = "FriendlyELEC";
+    name = "NanoPi M1";
+    identifier = "friendlyElec-nanoPiM1";
+    productPageURL = "https://wiki.friendlyelec.com/wiki/index.php/NanoPi_M1";
+  };
+
+  hardware = {
+    soc = "allwinner-h3";
+    allwinner.crust.defconfig = "nanopi_m1_defconfig";
+  };
+
+  Tow-Boot = {
+    defconfig = "nanopi_m1_defconfig";
+  };
+}


shvetsnikita:
Just in case, display issue is not specific to NanoPi M1, and also reproducible on OrangePi H3.

boards/friendlyElec-nanoPiM1/default.nix
```diff
@@ -0,0 +1,19 @@
+{ lib, ... }:
+
+{
+  device = {
+    manufacturer = "FriendlyELEC";
+    name = lib.mkDefault "NanoPi M1";
+    identifier = lib.mkDefault "friendlyElec-nanoPiM1";
+    productPageURL = "http://nanopi.io/nanopi-m1.html";
```
samueldr:
```suggestion
    productPageURL = "https://wiki.friendlyelec.com/wiki/index.php/NanoPi_M1";
```

While the product page is usually preferred, here we'll go for the wiki page, as the product page is from the site previous to the site previous to what is now FriendlyElec. AFAICT it's not sold anymore, so it makes sense to link there instead of the page that is starting to lose functionality.

boards/friendlyElec-nanoPiM1/default.nix
```diff
@@ -0,0 +1,19 @@
+{ lib, ... }:
+
+{
+  device = {
+    manufacturer = "FriendlyELEC";
+    name = lib.mkDefault "NanoPi M1";
+    identifier = lib.mkDefault "friendlyElec-nanoPiM1";
```
samueldr: Using `lib.mkDefault` here would only be useful if you intend to have other boards `import` this as a base device.

For reference:

 - https://github.com/Tow-Boot/Tow-Boot/blob/79acaaba7f44f91c0e0e32cf9062f2cb443575e6/boards/radxa-RockPi4/default.nix#L6-L7
 - https://github.com/Tow-Boot/Tow-Boot/blob/79acaaba7f44f91c0e0e32cf9062f2cb443575e6/boards/radxa-RockPi4C/default.nix#L2-L4

And it only makes sense for boards that are a follow-up of a given one.

If it is planned for the future, you can add `lib.mkDefault` when contributing the additional boards.

```suggestion
    name = "NanoPi M1";
    identifier = "friendlyElec-nanoPiM1";
```

boards/friendlyElec-nanoPiM1/default.nix
```diff
@@ -0,0 +1,19 @@
+{ lib, ... }:
+
```
samueldr: (See follow-up; since `lib` won't be needed, we can drop it.)

boards/friendlyElec-nanoPiM1/default.nix
```diff
@@ -0,0 +1,19 @@
...
+    productPageURL = "http://nanopi.io/nanopi-m1.html";
+  };
+
+  hardware = {
+    soc = "allwinner-h3";
+    allwinner.crust.defconfig = "nanopi_m1_defconfig";
+  };
+
+  Tow-Boot = {
+    defconfig = lib.mkDefault "nanopi_m1_defconfig";
```
samueldr:
```suggestion
    defconfig = "nanopi_m1_defconfig";
```

Same rationale as above.

samueldr (changes requested):
Hi!

Thanks for the contribution!

Looks mostly right. Things I shared here are more nits than outright issues, but should be fixed.

If you can, please amend and force push a new single commit; if you can't, no worries, I can handle that for the merge.

shvetsnikita:
> if you can't, no worries, I can handle that for the merge.

That would be helpful, thanks in advance.

One more question: is it okay to submit two other very similar boards that I don't have on hands? I talking about NanoPi M1 **Plus** and Libre Computer ALL-H3-CC **H3**. Chances of anything going wrong with these two is very low as I understand.

shvetsnikita:
Forgot to mention that I used lib.mkDefault because I intended to submit M1 Plus too:

```
{
  imports = [
    ../friendlyElec-nanoPiM1
  ];

  device = {
    manufacturer = "FriendlyELEC";
    name = "NanoPi M1 Plus";
    identifier = "friendlyElec-nanoPiM1Plus";
    productPageURL = "https://www.friendlyelec.com/index.php?route=product/product&path=69&product_id=176";
  };

  Tow-Boot = {
    defconfig = "nanopi_m1_plus_defconfig";
  };
}
```

samueldr:
> One more question: is it okay to submit two other very similar boards that I don't have on hands? I talking about NanoPi M1 Plus and Libre Computer ALL-H3-CC H3. Chances of anything going wrong with these two is very low as I understand.

It is required for the builds to have been verified to work, so you can submit it a as a draft PR by clearly stating that they have not been tested and testing that testing by owners of the board is welcome. Even though I agree that they are very likely to work, considering it's Allwinner.

shvetsnikita:
> It is required for the builds to have been verified to work, so you can submit it a as a draft PR by clearly stating that they have not been tested and testing that testing by owners of the board is welcome. Even though I agree that they are very likely to work, considering it's Allwinner.

Understood. What to do with lib.mkDefault here then? Will you remove it or keep it?

samueldr:
Force-pushsed:

 - The rebase on current `development`
 - Squashing the four fixups

samueldr (approved):

PR #264, opened 2023-09-06
odroid-HC4: Init board

erdnaxe:
The Hardkernel Odroid HC4 is similar to the Odroid C4.

There are however two strange behaviors:

1. Without calling `run boot_pci_enum` before `usb start`, the USB won't work.
2. `spi.installer.img` is unable to access the SPI. This bug is explained here: https://github.com/armbian/build/blob/4db330178fbe93b52e7987e8ce69ab9eaddd5377/patch/u-boot/v2023.01/board_odroidhc4/board.odroidhc4.hc4_sd_defconfig.for.writing.to.mtd.patch

I am unable to figure out how to configure Two-Boot to override `CONFIG_DEFAULT_DEVICE_TREE="meson-sm1-odroid-c4"` only for `spi.installer.img` (to circumvent the second strange behavior).

### DIFF
--- a/boards/odroid-HC4/default.nix
+++ b/boards/odroid-HC4/default.nix
@@ -1,0 +1,29 @@
+{ lib, pkgs, ... }:
+
+{
+  device = {
+    manufacturer = "Hardkernel";
+    name = "ODROID-HC4";
+    identifier = "odroid-HC4";
+    productPageURL = "https://www.hardkernel.com/shop/odroid-hc4/";
+  };
+
+  hardware = {
+    soc = "amlogic-s905x3";
+    SPISize = 16 * 1024 * 1024; # 16 MiB
+  };
+
+  Tow-Boot = {
+    defconfig = "odroid-hc4_defconfig";
+    config = [
+      (helpers: with helpers; {
+        USE_PREBOOT = yes;
+        # 'run boot_pci_enum' is required before 'usb start' to have working USB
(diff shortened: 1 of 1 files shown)


erdnaxe:
I flashed `nixos-minimal-23.05.3242.da5adce0ffaf-aarch64-linux.iso` to an USB drive and tried to boot on it.
I see the NixOS boot menu, then it starts booting the kernel and freezes on:
```
usb usb1-port2: unable to enumerate USB device
/dev/root: Can't open blockdev
/dev/root: Can't open blockdev
/dev/root: Can't open blockdev
```

Could this caused by a bad device-tree in U-Boot?

erdnaxe:
Things tried:
- flash `nixos-minimal-new-kernel-23.05pre456191.b69883faca9-aarch64-linux.iso` on USB disk: **no effects**
- rebase on https://github.com/Tow-Boot/Tow-Boot/pull/255: **no effects**
- add a patch (in `Tow-Boot.patches`) to update meson-g12-common.dtsi, meson-sm1-odroid-hc4.dts, meson-sm1-odroid.dtsi and meson-sm1.dtsi to Linux 6.5 version: **no effects**

It seems that when booting Linux kernel, the USB bus loses power.

From what I understand, U-Boot gives a device-tree to the kernel. Is there a simple way to let the NixOS kernel load its own device-tree (for debug purposes)?

PR #266, opened 2023-10-07
Add support for FriendlyELEC NanoPi K1 Plus

shvetsnikita:
Hello!

This is work in progress. I tried to add NanoPi K1 Plus board support, but build fail for some reason:

```
***
*** Can't find default configuration "./configs/nanopi_k1_plus_defconfig"!
***
make: *** [3rdparty/kconfig/Makefile:49: nanopi_k1_plus_defconfig] Error 1
error: builder for '/nix/store/qanzsx3bn36hdwahlw6rc21g9z5gwzyb-crust-firmware-untagged-2023-02-28.drv' failed with exit code 2
error: 1 dependencies of derivation '/nix/store/3r51blna4kv3z1a797ilb9h5il521r7j-Tow-Boot-nanopi_k1_plus_defconfig-noenv-aarch64-unknown-linux-gnu-2023.07-007-pre.drv' failed to build
error: 1 dependencies of derivation '/nix/store/5l90rm73n1sbxr4rn1njkhqvqiapysjq-Tow-Boot.friendlyElec-nanoPiK1Plus.2023.07-007-pre.drv' failed to build

```

### DIFF
--- a/boards/friendlyElec-nanoPiK1Plus/0001-nanopi-k1-plus-board-enablement.patch
+++ b/boards/friendlyElec-nanoPiK1Plus/0001-nanopi-k1-plus-board-enablement.patch
@@ -1,0 +1,172 @@
+diff --git a/arch/arm/dts/Makefile b/arch/arm/dts/Makefile
+index b6eebe8..a6eb75b 100644
+--- a/arch/arm/dts/Makefile
++++ b/arch/arm/dts/Makefile
+@@ -371,6 +371,7 @@ dtb-$(CONFIG_MACH_SUN50I_H5) += \
+ 	sun50i-h5-orangepi-zero-plus.dtb \
+ 	sun50i-h5-nanopi-m1-plus2.dts \
+ 	sun50i-h5-orangepi-pc2.dtb \
++	sun50i-h5-nanopi-k1-plus.dtb \
+ 	sun50i-h5-orangepi-prime.dtb \
+ 	sun50i-h5-orangepi-zero-plus2.dtb
+ dtb-$(CONFIG_MACH_SUN50I) += \
+diff --git a/arch/arm/dts/sun50i-h5-nanopi-k1-plus.dts b/arch/arm/dts/sun50i-h5-nanopi-k1-plus.dts
+new file mode 100644
+index 0000000..c08af78
+--- /dev/null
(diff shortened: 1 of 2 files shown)


shvetsnikita:
Source: https://github.com/armbian/build/tree/main/patch/u-boot/u-boot-sunxi/board_nanopik1plus

samueldr:
Right, this looks harder than it actually is.

```
error: builder for '/nix/store/qanzsx3bn36hdwahlw6rc21g9z5gwzyb-crust-firmware-untagged-2023-02-28.drv' failed with exit code 2
                                                                ^^^^^
```

The reason ***crust*** fails to build is because it's implicitly referring to a non-existent config:

 - https://github.com/Tow-Boot/Tow-Boot/blob/f4d3d1dd344c3f2845f5b65427ddd6625f66177d/modules/hardware/allwinner/default.nix#L63-L65
 - https://github.com/crust-firmware/crust/tree/master/configs

Crust is not *required* for Allwinner systems.

Though it is generally assumed to exist. When not available, for now we use this unergonomic way to mark it as unavailable

```
diff --git a/boards/friendlyElec-nanoPiK1Plus/default.nix b/boards/friendlyElec-nanoPiK1Plus/default.nix
index eeae93f..ba096d7 100644
--- a/boards/friendlyElec-nanoPiK1Plus/default.nix
+++ b/boards/friendlyElec-nanoPiK1Plus/default.nix
@@ -15,5 +15,8 @@
     patches = [
       ./0001-nanopi-k1-plus-board-enablement.patch
     ];
+    builder.additionalArguments = {
+      SCP = "/dev/null";
+    };
   };
 }
```

This is because, somewhat by design, we have to dip ourselves into the implementation details in U-Boot to mark it as non-existent, as it is really preferred to have it available.

With this change Tow-Boot should build (it built for me).

shvetsnikita:
Indeed it helps, thanks!

Although I intentionally avoided [this patch](https://github.com/armbian/build/blob/main/patch/u-boot/u-boot-sunxi/board_nanopik1plus/allwinner-disable-de2-to-improve-edid-detection.patch), there is nothing on the screen until KMS initialization. I tried to specify CONFIG_VIDEO_DE2=y but it didn't help. Any ideas what else worth a shoot? 

I assume as is (no HDMI output at bootloader stage) it can not be merged? Or there is value in ability to select boot device at least via serial console? This board support eMMC.

PR #270, opened 2023-10-08
Tag output with uSWIDs

### LINKED ISSUE #129: Version reporting
Would be great if the version is easily retrievable from the firmware.

Current options:
- DMI property (UEFI-only), which would allow us to use standardized interfaces, so my preference.
- Adding a string `VERSION=2022.01-004`
- Modifying the reported version by dropping the date: `Tow-Boot 2022.01-004`

Would help us for https://github.com/fwupd/fwupd/issues/4294

samueldr:
Fixes #129

* * *

With these changes, we now append an uSWID to the actual Tow-Boot "binaries", which in turn should unblock future work around fwupd.

```
~/.../Tow-Boot/uswid $ ./result-uswid/bin/uswid --load result/binaries/Tow-Boot.mmcboot.bin --save uswid.xml
Found USWID header at offset: 817665
```

```xml
<?xml version='1.0' encoding='utf-8'?>
<SoftwareIdentity xmlns="http://standards.iso.org/iso/19770/-2/2015/schema.xsd" xmlns:SHA256="http://www.w3.org/2001/04/xmlenc#sha256" xmlns:SHA512="http://www.w3.org/2001/04/xmlenc#sha512" xmlns:n8060="http://csrc.nist.gov/ns/swid/2015-extensions/1.0" xml:lang="en-US" name="Tow-Boot for Pinebook (A64) [variant: mmcboot]" tagId="qnhl8243ns10vpcg0x1nbx7dwp3p0gjb-Tow-Boot-pinebook_defconfig-mmcboot-aarch64-unknown-linux-gnu-2023.07-007-rc1-pre" version="2023.07-007-rc1-pre" versionScheme="multipartnumeric+suffix">
  <Entity name="Tow-Boot" regid="tow-boot.org" role="tagCreator licensor maintainer softwareCreator"/>
  <Meta product="Tow-Boot" persistentId="org.tow-boot.tow-boot"/>
</SoftwareIdentity>
```

And the ini from the uSWID:

```ini
[uSWID]
tag-id = qnhl8243ns10vpcg0x1nbx7dwp3p0gjb-Tow-Boot-pinebook_defconfig-mmcboot-aarch64-unknown-linux-gnu-2023.07-007-rc1-pre
software-name = Tow-Boot for Pinebook (A64) [variant: mmcboot]
software-version = 2023.07-007-rc1-pre
version-scheme = multipartnumeric+suffix
summary = An opinionated distribution of U-Boot
product = Tow-Boot
persistent-id = org.tow-boot.tow-boot

[uSWID-Entity:TagCreator]
name = Tow-Boot
regid = tow-boot.org
extra-roles = Licensor,Maintainer,SoftwareCreator
```

* * *

cc @DylanVanAssche, @hughsie, especially regarding fwupd, is there anything outright problematic in here?

I've used [NISTIR 8060 Guidelines for the Creation of Interoperable Software Identification (SWID) Tags](https://nvlpubs.nist.gov/nistpubs/ir/2016/NIST.IR.8060.pdf) as a source of truth to know what tags meant, and their *required* state or not. But I don't assume I've understood all and how it applies in practice here.

<del>(Note that the empty *evidence* values are currently a workaround, see https://github.com/hughsie/python-uswid/issues/53.)</del>

Now, what's up with that weird `tag-id`?? Glad you asked! It's the basename of the Nix store path for the build of the binary. They are a hash encapsulating the whole derivation, which includes the whole build environment. Hopefully that is not a problematic `tag-id` to use, as not only is it convenient, it is actively unique to the whole stack used to produce the build *including what was used to produce the uSWID*.

### About git revision

Adding the git revision used for the build *could* be done, but is kind of unergonomic when referring to the repository in which the expressions reside. The best way to do it would be "from the outside" when e.g. building the project, passing `--argstr rev "$(...)"`.

It might be worth it to add more useful information for the build, but would be reserved, I suppose, for release builds mainly, and is really orthogonal to these changes. It wouldn't be hard to add it in the SWID after the fact.

### DIFF
--- a/modules/tow-boot/builder.nix
+++ b/modules/tow-boot/builder.nix
@@ -19,8 +19,6 @@ let
     version = config.Tow-Boot.uBootVersion;
     structuredConfig = (config.Tow-Boot.structuredConfigHelper version);
   };
-
-  towBootIdentifier = "${config.Tow-Boot.releaseNumber}${config.Tow-Boot.releaseIdentifier}";
 in
 {
   options = {
@@ -240,6 +238,7 @@ in
           uBootVersion
           outputName
           buildUBoot
+          towBootIdentifier
         ;
         inherit (config.Tow-Boot.builder)
           additionalArguments
@@ -250,7 +249,6 @@ in
           postPatch
         ;
         boardIdentifier = config.device.identifier;
-        inherit towBootIdentifier;
       });
 
       builder = {
--- a/modules/tow-boot/builder.nix
+++ b/modules/tow-boot/builder.nix
@@ -98,6 +98,7 @@ in
         , buildInputs
         , nativeBuildInputs
         , postPatch
+        , uswidHelper
         }:
 
         stdenv.mkDerivation ({
@@ -222,6 +223,30 @@ in
             cp .config $out/config/$variant.config
             mkdir -p $out/binaries
             ${installPhase}
+            if test -e $out/binaries; then
+              (
+              echo ":: Adding uSWID data"
+              cd $out/binaries
+              for binary in "${outputName}"*.bin; do
+                printf " - '%s' \n" "$binary"
+                ${/*
+                  Most builds should produce only one output here, but in case other outputs
+                  grow, we'll pre-empt problems by appending the binary name to the tagId.
+
+                  We're referring to tagId as ISO/IEC 19770-2:2015 implies: “a unique
+                  reference for the **specific** […] binary […] If two tagIDs match […] the
+                  underlying products they represent are […] exactly the same”.
+
+                  **Always** treat this tagId as an opaque blob. The useful data it
+                  contains should be found elsewhere.
+                */""}
+                tagId="$(echo $(basename $out)/$binary | sed -e 's/-${stdenv.cc.targetPrefix}/-/')"
+                ${uswidHelper} "$tagId" --compress --save uswid.bin
+                cat uswid.bin >> "$binary"
+                rm uswid.bin
+              done
+              )
+            fi
             runHook postInstall
           '';
 
@@ -249,6 +274,7 @@ in
           postPatch
         ;
         boardIdentifier = config.device.identifier;
+        uswidHelper = config.Tow-Boot.uswid.output.helper;
       });
 
       builder = {
--- a/modules/tow-boot/uswid.nix
+++ b/modules/tow-boot/uswid.nix
@@ -1,0 +1,128 @@
+{ config, lib, pkgs, ... }:
+
+let
+  inherit (lib)
+    concatStringsSep
+    mkOption
+    optional
+    types
+  ;
+
+  inherit (config.Tow-Boot)
+    outputName
+    towBootIdentifier
+    uBootVersion
+    variant
+  ;
+
+  inherit (config)
+    device
+  ;
+
+  #
+  # For the definition of the fields, refer to:
+  #
+  #  - ISO/IEC 19770-2:2015 (Preferred)
+  #      - https://standards.iso.org/iso/19770/-2/2015/schema.xsd
+  #      - https://web.archive.org/web/20230824211322/https://standards.iso.org/iso/19770/-2/2015/schema.xsd
+  #
+  #  - RFC9393 (Adequate)
+  #      - https://datatracker.ietf.org/doc/html/rfc9393
+  #
+  #  - NIST.IR 8060 (Okay, too)
+  #      - https://nvlpubs.nist.gov/nistpubs/ir/2016/NIST.IR.8060.pdf
+  #
+  uSWIDData = {
+    "uSWID" = {
+      # Will be replaced during execution with basename of $out (guaranteed meaningfully unique)
+      # DO NOT assume any useful form when consuming the SWID.
+      tag-id = "$id";
+
+      # (Unchanging) meaningful name for end-user
+      software-name = "${outputName} for ${device.name} [variant: ${variant}]";
+
+      # Short description for the software (shown in e.g. fwupd frontends)
+      summary = "An opinionated distribution of U-Boot";
+
+      # Lexicographically sortable version name
+      # (pre/post suffixes will not sort as expected, but that is okay)
+      software-version = "${"${uBootVersion}-${towBootIdentifier}"}";
+      version-scheme = "multipartnumeric+suffix";
+
+      # Tow-Boot or U-Boot, depending on how the tooling is used.
+      product = outputName;
+
+      # An identifier (GUID suggested) that refers to a "set of software components that are related",
+      # "but may be different versions".
+      #
+      # We're using the device identifier and variant here, since the only "related"
+      # components would be the same board, same variant components here.
+      # Changing from one variant to another is a "major" change all things considered.
+      #
+      # When tracking updates, assuming the same `persistent-id` works on the same system is okay.
+      persistent-id = "org.Tow-Boot.${outputName}.${device.identifier}.${variant}";
+    };
+    # The entity producing this tag...
+    "uSWID-Entity:TagCreator" = {
+      name = "Tow-Boot";
+      regid = "tow-boot.org";
+      # ... And its role(s) with regard to the software.
+      # "pre-defined roles include: aggregator, distributor, licensor, softwareCreator, tagCreator"
+      extra-roles = concatStringsSep "," (
+        [
+          # Distributor assumed to mean "provides binaries"
+          "Distributor"
+          "Licensor"
+          "Maintainer"
+        ]
+        # Creator of the "distribution"
+        ++ (optional (outputName == "Tow-Boot") "SoftwareCreator")
+      );
+    };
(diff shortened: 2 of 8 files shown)


hughsie:
> They are a hash encapsulating the whole derivation, which includes the whole build environment

That sounds like you're re-inventing the other two hashes :) e.g. see https://github.com/hughsie/python-uswid#ini-file

The colloquial-version should be a hash of all the source files, and the edition should describe the "tree" -- i.e. anything build in the same tree would have the same edition, but change the source files within a tree and the colloquial-version should be different.

I think in general the tag ID should be able to be compared *between* SWIDs -- i.e. the SWID from last year should have the same set of tag-id as the one from today.

DylanVanAssche:
From my side, it looks good, but @hughsie is a better expert on uSWID than me :P
As long as I can make fwupd happy with a good version number, it is fine for me.

samueldr:
First of all, thank you for taking the time in answering.

I believe there are two distinct issues here, so we don't talk past each-other, I'll title them in how I see them. Sorry if this is a bit long.

Also note that I'm not trying to contradict you, but trying to *understand fully* what I'm doing. I am not discounting that I could be missing something.

* * *

### Arbitrary reproducible `tag-id` source

> > They are a hash encapsulating the whole derivation, which includes the whole build environment
> 
> That sounds like you're re-inventing the other two hashes :) e.g. see https://github.com/hughsie/python-uswid#ini-file

The goal here is not what I think you're understanding from my description.

It isn't to have it be the meaningful hashes, but be a unique predictable and reproducible value, in the meaning of [Reproducible builds](https://reproducible-builds.org/). Given the same inputs, it should be expected that we have the same outputs, so using e.g. a random GUID here breaks reproducibility.

**In that optic**, is using that arbitrary `tag-id` problematic?

My understanding is that it wouldn't be, as NIST.IR.8060 implies it can be anything that is globally unique as an identifier for the tagged product. *Assuming the binary is reproducible* in all other aspects (last I checked, the U-Boot builds were), it means that whoever builds the binary here gets the same binary, and that it would be correct to use a reproducible identifier here.

RFC9393 stats it uniquely references *a software component*. Here I'm assuming *software component* would refer to a build thereof.

That's why I'm using the store path basename, as it intrinsically is unique to the given inputs, and *assumed* to be producing the same output.

In other word, we get a globally unique identifier *for free* from the intrinsic properties of the build system, so is there any technical reason not to use them?

(Side-note: let's absolutely trust those assumptions about reproducibility in this discussion, whether or not they apply is irrelevant when asking *assuming those assumptions are right*.)


* * *

### Tags usage

I'm unsure what to make of the tags and their usage, seeing the web generator/uswid, and NIST.IR.8060 don't seem to agree about the semantics.

> That sounds like you're re-inventing the other two hashes :) [...]
> 
> The colloquial-version should be a hash of all the source files, and the edition should describe the "tree" -- i.e. anything build in the same tree would have the same edition, but change the source files within a tree and the colloquial-version should be different.

#### `quolloquial-version`

Quoting from NIST.IR.8060, for `<Meta>`:

> `@colloquialVersion`: the informal version of the product (e.g., 2013). The colloquial version may be the same through multiple releases of a software product where the `@version` specified in `<SoftwareIdentity>` is much more specific and will change for each software release.

From RFC9393:

> A textual value for the software component's informal or colloquial version. Examples may include a year value, a major version number, or a similar value used to identify a group of specific software component releases that are part of the same release/support cycle. This version can be the same through multiple releases of a software component, while the software-version specified in the concise-swid-tag group is much more specific and will change for each software component release. This version is intended to be used for string comparison (byte by byte) only and is not intended to be used to determine if a specific value is earlier or later in a sequence.

My understanding here, for Tow-Boot, is that I would be putting `2023.07-007` here. Whether it's `rc1`, `rc5`, the final release, an in-dev  version. It would be the same version commonly referred to when speaking in general terms.

#### `edition`

Quoting from NIST.IR.8060, for `<Meta>`:

> `@edition`: the variation of the product (e.g., “`Home`”, “`Enterprise`”, “`Student`”).

From RFC9393:

> A textual value indicating that the software component represents a functional variation of the code base used to support multiple software components. For example, this item can be used to differentiate enterprise, standard, or professional variants of a software component.

My understanding here is that *edition* would never change between builds, even when updated... A bad example would be making a "Standard" and "Pro" version of Tow-Boot, that's what would be put in Edition. So using any "input" like the git revision identifier isn't correct?

(I debated putting the variant as edition, but the variant is a not relevant as a "functional variation", it's more of an implementation detail. Relatedly, the device identifier maybe is more suited for the edition...)

#### `tag-id`

> I think in general the tag ID should be able to be compared _between_ SWIDs -- i.e. the SWID from last year should have the same set of tag-id as the one from today.

I'm not sure I follow this. What does "last year" refer to here? Last year when building the same source tree? They will be. There are no untracked inputs (nor impurities) here that would change the hash, nor the outputs.

If it refers to a different release of the software, does that imply that every different builds will have the same identifier? That breaks referring to a specific "software component"'s SWID (a binary's SWID) using `swid:${tag-id}`, as it would refer to many other too.

RFC9393 seem to agree with my interpretation here.

hughsie:
Woah, lots to unpack there. I'm just trying to tell you what I'm advising they use the "colloquial-version" and "edition" for -- it's not according to the specifications, as I'm trying to be pragmatic. It's only advice, you can ignore it if you want to.

samueldr:
Yeah, sorry, if that was a lot.

I was trying to understand where the advice comes from, since it didn't map with what I've found about SWIDs. Again, I'm trying to ensure I understand what I'm implementing fully here, since I believe this is a crucial thing to get right.

I don't want to find out later down the line that either something doesn't match SWID (whatever that means), or that it breaks assumptions under LVFS/fwupd.

Assuming my understanding of SWIDs is correct, are there problems under the assumptions of LVFS/fwupd with using those values as they are specced?

Or maybe to frame it in another way: is there a "SWIDs as actually found in the wild" guide book that might help here?

And again, thank you for taking with this.

hughsie:
> is there a "SWIDs as actually found in the wild" guide book

We wrote this if that helps: https://uefi.org/node/4424

modules/tow-boot/builder.nix
```diff
@@ -224,6 +223,30 @@ in
...
+                ${/*
+                  Most builds should produce only one output here, but in case other outputs
+                  grow, we'll pre-empt problems by appending the binary name to the tagId.
+
+                  We're referring to tagId as ISO/IEC 19770-2:2015 implies: “a unique
+                  reference for the **specific** […] binary […] If two tagIDs match […] the
+                  underlying products they represent are […] exactly the same”.
+
+                  **Always** treat this tagId as an opaque blob. The useful data it
+                  contains should be found elsewhere.
```
samueldr, line 241: Reasoning for the choice of `tagId` identifier here.

modules/tow-boot/uswid.nix
```diff
@@ -0,0 +1,128 @@
...
+
+      # (Unchanging) meaningful name for end-user
+      software-name = "${outputName} for ${device.name} [variant: ${variant}]";
+
+      # Short description for the software (shown in e.g. fwupd frontends)
+      summary = "An opinionated distribution of U-Boot";
+
+      # Lexicographically sortable version name
+      # (pre/post suffixes will not sort as expected, but that is okay)
+      software-version = "${"${uBootVersion}-${towBootIdentifier}"}";
```
samueldr, line 49: This currently looks like:

```
2023.07-007-rc2-pre
```

Actual releases will look like:

```
2023.07-007
```

There is no equivalent to a "patch" release of Tow-Boot... If a release needs to happen quickly, our release number (007 here) is increased. The prefix will always increase, and follows the upstream (U-Boot) versioning scheme.

modules/tow-boot/uswid.nix
```diff
@@ -0,0 +1,128 @@
...
+
+      # An identifier (GUID suggested) that refers to a "set of software components that are related",
+      # "but may be different versions".
+      #
+      # We're using the device identifier and variant here, since the only "related"
+      # components would be the same board, same variant components here.
+      # Changing from one variant to another is a "major" change all things considered.
+      #
+      # When tracking updates, assuming the same `persistent-id` works on the same system is okay.
+      persistent-id = "org.Tow-Boot.${outputName}.${device.identifier}.${variant}";
```
samueldr, line 63: For e.g. the SPI build for the *La Frite* SBC, it would look like this:

```
org.Tow-Boot.Tow-Boot.libreComputer-amlS805xAc.spi
                      ^^^^^^^^^^^^^^^^^^^^^^^^ ---
```

The expectation is that software consuming the uSWID can determine a build is for the same board *and* the same installation method using `persistent-id` value. If it is equal, it is "the same", but another version or edition. In other words: is updated using the same semantics for a given board+installation method combination.

samueldr (commented):
I've added further explanation/reasoning attached to the values used here. I believe that should be correct, and all that is needed for tracking updates.

PR #280, opened 2023-10-18
Tow-Boot: update to -rc4 (rockchip + pinephonePro)

samueldr:
I'm misusing this to also handle https://github.com/Tow-Boot/U-Boot/pull/3

Fixes #297

* * *

Apparently I left a bunch of the work incomplete for the Pinephone Pro, as I hadn't verified that stuff like the buttons, LED and vibrator had their bindings.

With this change set, only those critical parts have been ported forward. We'll try not to stray too far from the upstream DT, and with the inevitable sync with mainline, we'll try to make it work even better.

Also adds back a change that removes an `iodomain` change, which breaks SPI Flash in e.g. the installer.

Finally, update a comment so I know why those config changes are required.

 - https://github.com/samueldr/u-boot/compare/tb-2023.07-007-rc3..wip/tb2307/ppp-dt

### What was done

 - Tested on Pinephone Pro, buttons, vibration and LED pattern started working again.

There is no difference for any other platform.

* * *

### DIFF
--- a/boards/pine64-pinephonePro/default.nix
+++ b/boards/pine64-pinephonePro/default.nix
@@ -44,7 +44,9 @@
         CMD_POWEROFF = lib.mkForce yes;
       })
       (helpers: with helpers; {
-        # Workarounds required for eMMC issues and current patchset.
+        # NOTE: works around an issue with the config, where we hit:
+        #     mmc fail to send stop cmd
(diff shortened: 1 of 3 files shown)


L-as:
I tested this and it works fine, but I couldn't seem to boot from the SD card, albeit UMS works.

Issue #289, opened 2023-12-30
Can't enter tb_menu over serial console by pressing ESCAPE or CTRL-C  on OG PP


AndreySV:
It's regression, it worked on previous Tow-Boot release. 
Hardware OG PP, Software Tow-Boot 2023.07 007-rc3-pre [variant: mmcboot]

Serial output hangs for couple seconds before 
```
Core:  75 devices, 21 uclasses, devicetree: separate
WDT:   Not starting watchdog@1c20ca0
MMC:   mmc@1c0f000: 0, mmc@1c10000: 2, mmc@1c11000: 1
Loading Environment from nowhere... OK
In:    serial
Out:   serial
Err:   serial
Net:   No ethernet found.
Please press [ESCAPE] or [CTRL+C] to enter the boot menu.

```
and then jumps to U-Boot menu.

AndreySV:
According to the code, it looks like expected behavior for phones.
https://github.com/Tow-Boot/Tow-Boot/blob/development/modules/tow-boot/phone-ux.nix#L198
If someone needs access to u-boot cmdline for development, has to built custom Tow-Boot with non-zero BOOTDELAY.

PR #290, opened 2023-12-30
modules/tow-boot: Fix missing bootdelay argument in AUTOBOOT_PROMPT

AndreySV:
printf(CONFIG_AUTOBOOT_PROMPT, bootdelay) takes bootdelay argument to print the value upon bootup. CONFIG_AUTOBOOT_PROMPT includes the %d for all boards. Add %d to Tow-Boot AUTOBOOT_PROMPT.

Additionally that would help to understand, why it's not possible to enter tb_menu using keys on a phone variant over UART. [1]

1. https://github.com/Tow-Boot/Tow-Boot/issues/289

### DIFF
--- a/modules/tow-boot/config.nix
+++ b/modules/tow-boot/config.nix
@@ -60,7 +60,7 @@ in
           bright = "\\e[1m";
         in
         lib.mkIf (!config.Tow-Boot.buildUBoot) (
-          freeform ''"${reset}Please press [${bright}ESCAPE${reset}] or [${bright}CTRL+C${reset}] to enter the boot menu."''
+          freeform ''"${reset}Please press [${bright}ESCAPE${reset}] or [${bright}CTRL+C${reset}] to enter the boot menu in %ds."''
         )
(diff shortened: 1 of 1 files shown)


samueldr:
Hi, again sorry it took some time to get to this.

I'll have to test, maybe 2023.07-007 broke that in some way, but with serial I was able to just mash escape before the timeout was shown, and it would get picked up.

In other words, in `-006` at the very least, IIRC, the input buffer wasn't used-up before the check, and 0 would still end-up running the check once.

Though, do I understand correctly that this change **does not** fix the issue with the phone UX?

AndreySV:
Thanks, for merging the change.

yes, it doesn't fix the issue with Phone UX. I was thinking, that no-way to enter tb_menu over serial console is expected behavior for phones.

This MR fixes only CONFIG_AUTOBOOT_PROMPT usage with printf. See. https://github.com/lentinj/u-boot/blob/master/doc/README.autoboot

PR #291, opened 2023-12-30
arm-trusted-firmware: 2.9 -> 2.10

AndreySV:
That will bring-in warm reset feature on Sunxi A64 platform (including PinePhone). It's needed to enable persistent storage support (pstore) in Linux kernel to get crash logs after reset.


samueldr:
Sorry it took a while to get to this, I verified it all still builds. Not verified the results, but given the project and how it operates, there is no reason to think there will be an issue. If there is, it will be picked-up when testing later on.

Thank you!

PR #294, opened 2024-01-05
crust-firmware: untagged-2023-02-28 -> 0.6

AndreySV:
It would be nice to include in a release tagged (released) version of the program.


samueldr:
Sorry it took a while to get to this, I verified it all still builds. Not verified the results, but given the project and how it operates, there is no reason to think there will be an issue. If there is, it will be picked-up when testing later on.

Thank you!

PR #296, opened 2024-01-19
Add support for FriendlyElec NanoPi R4S

arnarg:
FriendlyElec NanoPi R4S is an SBC based on RK3399.

The standard model has no eMMC (there's an "enterprise" edition with this, which I do not have) nor SPI flash (although SPI1 is available on a small GPIO header, so it can be added).

What's tested:

- Tow-Boot is flashed to shared storage on SD card.
- Console is available on serial header.
- UEFI boot with NixOS 23.11 USB ISO.

### DIFF
--- a/boards/friendlyElec-nanoPiR4S/default.nix
+++ b/boards/friendlyElec-nanoPiR4S/default.nix
@@ -1,0 +1,17 @@
+{
+  device = {
+    manufacturer = "FriendlyELEC";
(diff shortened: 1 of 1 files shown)


samueldr:
Hi!

Sorry it took a while to get back to this. At a glance this builds, no SPI Flash listed on the product page, and given your previous testing and general RK3399 support in U-Boot, should be good.

Thanks!

PR #297, opened 2024-01-22
rk3399: move ramdisk_addr_r for larger kernels

artemist:
The NixOS Linux 6.7 kernel is too big to fit in the 63.5MiB of space allocated in the U-Boot config. This temporary fix adds an additional 32MiB of space for the kernel.

I tested this on a PineBook Pro but do not have any other RK3399 devices to test. It would be useful of U-Boot didn't hardcode initrd position but that would be a much more invasive change than this.


samueldr (changes requested):
(Note to self to not merge this)

Replaced by merging this in the rk3399 branch:

 - https://github.com/Tow-Boot/U-Boot/pull/3

This is technically approved, and known to be needed.

PR #340, opened 2024-11-03
docker/build: Fix `Permission denied` error when building with Docker shim on MacOS

### LINKED ISSUE #339: `Permission denied` error when building with Docker shim on MacOS
On MacOS if you try to build with Docker shim the `release-2023.07-007` tag, you get the following error:
```
> ./support/docker/build.sh -A uBoot-sandbox
<unimportant output>

(Unwrapping store paths...)
result -> /nix/store/h0pa6p29lnnvfx4m4v0713syzfg1jfyy-Tow-Boot-sandbox_defconfig-noenv-2023.07-007
cp: setting permissions for '/Tow-Boot/result': Permission denied
```

After that I have the `./result` directory created with the `dr-xr-xr-x` permissions, but it has nothing inside.

I guess, the issue is somewhere in between MacOS FS permissions, and the intermediary Linux VM, that `Docker Desktop` for Mac has to use, to run actual containers inside. And `cp` reporting error, because it can't copy files into `result` directory it just created, because it has no Write permissions.

glukki:
With Docker on Mac `cp -R` fails to copy anything into `result` directory after creating it without write permissions.
This change makes a build result copy within container and adds write permissions to the files inside, before writing them into the `/Tow-Boot/` directory.

Please try building with shim on Linux/Windows

Fixes #339

### DIFF
--- a/support/docker/build.sh
+++ b/support/docker/build.sh
@@ -74,7 +74,7 @@ if [ -e /Tow-Boot ]; then
 		cp -r "$target" "$proxy_dir/$f"
 		chmod -R u+w "$proxy_dir/$f"
 
-		cp -r "$proxy_dir/$f" "$current_dir/$f"
+		mv "$proxy_dir/$f" "$current_dir/$f"
 		chown -R "$REAL_UID:$REAL_GID" "$current_dir/$f"
 	done
 else
--- a/support/docker/build.sh
+++ b/support/docker/build.sh
@@ -61,11 +61,20 @@ if [ -e /Tow-Boot ]; then
 
 	stderr.printf "\n(Unwrapping store paths...)\n"
 
+	proxy_dir="/tmp-proxy-dir"
+	mkdir -p "$proxy_dir"
+
 	for f in *; do
 		target="$(readlink -f "$f")"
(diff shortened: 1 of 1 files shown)


support/docker/build.sh
```diff
@@ -61,11 +61,20 @@ if [ -e /Tow-Boot ]; then
...
+
 	for f in *; do
 		target="$(readlink -f "$f")"
 		rm -rf "$current_dir/$f"
 		stderr.printf "%s -> %s\n" "$f" "$target"
-		cp -r "$target" "$current_dir/$f"
+
+		# Docker on Mac fails to copy into read-only `result` directory
+		# So we copy into proxy directory within container and add write permissions first
+		cp -r "$target" "$proxy_dir/$f"
```
glukki, line 74: Is it worth running `rm -r "$proxy_dir"` in the end, to reduce the new container layer size?

samueldr: Or instead of (later) `cp`'ing the files, would `mv`'ing the files from it handle that gracefully?

glukki: updated

